PR #20, opened 2022-10-26
Add creator (un)verify support (and collection)

alexandruag:
Add basic support for ingesting creator (un)verify events.  This is currently on top of a previous `main`, as rebasing on top of the latest one leads to some issues. Would be great to sanity check and review the approach while those get worked out. Will rebase and update the PR when everything fits together properly.

### DIFF
--- a/das_api/Cargo.lock
+++ b/das_api/Cargo.lock
@@ -469,9 +469,8 @@ checksum = "8d696c370c750c948ada61c69a0ee2cbbb9c50b1019ddb86d9317157a99c2cae"
 
 [[package]]
 name = "blockbuster"
-version = "0.5.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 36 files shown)


alexandruag:
Just a quick heads-up that I'm going to either update this PR or create a new one to rebase and include collection stuff as well. Also, ever since the container deployment upgraded to a more recent version of `digital-asset-validator-plugin`, the weirdness where blockbuster was receiving accounts in a strange order is gone, and now we can access them in the same order as they appear in the anchor struct definitions proper.

alexandruag:
Pushed some changes for collection verification as well. For the immediate time being they are mostly to showcase the approach, as we need to merge https://github.com/metaplex-foundation/blockbuster/pull/16, and then perform a rebase to have this in merging shape. I'm also investigating an issue that happens if the code from one of the temp commits is not commented, but that shouldn't take long.

alexandruag:
Ready for review

PR #22, opened 2022-10-27
Danenbm/backfiller retries

danenbm:
### Notes
* Add failed column to backfill table.
* Modified backfiller to retry on failure but give up after max tries.
* Also added some updates to the README.

### Testing
* Injected a failure in the backfilling process and observed backfiller try 5 times to backfill a tree, then give up and mark the tree as failed to backfill.
* Injected an extra row to kick off a normal backfill, and observed successful backfill process (albeit skipping RPC stuff because it was a fake extra row).  I was just looking for regression on the overall triggering process.

### DIFF
--- a/digital_asset_types/src/dao/generated/backfill_items.rs
+++ b/digital_asset_types/src/dao/generated/backfill_items.rs
@@ -18,9 +18,9 @@ pub struct Model {
     pub tree: Vec<u8>,
     pub seq: i64,
     pub slot: i64,
-    pub force_chk: Option<bool>,
-    pub backfilled: Option<bool>,
-    pub failed: Option<bool>,
+    pub force_chk: bool,
+    pub backfilled: bool,
+    pub failed: bool,
 }
 
 #[derive(Copy, Clone, Debug, EnumIter, DeriveColumn)]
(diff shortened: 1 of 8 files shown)


digital_asset_types/src/dao/generated/backfill_items.rs
```diff
@@ -20,6 +20,7 @@ pub struct Model {
     pub slot: i64,
     pub force_chk: Option<bool>,
     pub backfilled: Option<bool>,
+    pub failed: Option<bool>,
```
alexandruag, line 23: Just wondering, why is this (and maybe the others as well) an `Option<bool>` instead of just a `bool`?

alexandruag (approved):
LGTM, just left a small question

digital_asset_types/src/dao/generated/backfill_items.rs
danenbm: The answer is that SeaORM uses Option when the database field can also be NULL.

However, this question made me think that I don't need to allow NULL for any of the columns in `backfill_items`  so I updated the table and generated code in my latest commit.

danenbm: The answer is that SeaORM uses Option when the database field can also be NULL.

However, this question made me think that I don't need to allow NULL for any of the columns in `backfill_items`  so I updated the table and generated code in my latest commit.

alexandruag (approved):

PR #26, opened 2022-10-31
feat: expose more fields to fix decompress using asset from api call

stranzhay:
we need to return these additional fields so that the wallet providers can use them from api call response to call methods like redeem -> decompress correctly since they likely will not have access to original `MetadataArgs`


digital_asset_types/src/dapi/asset.rs
```diff
@@ -285,6 +281,43 @@ pub fn asset_to_rpc(asset: FullAsset) -> Result<RpcAsset, DbErr> {
             total: u.get("total").and_then(|t| t.as_u64()).unwrap_or(0),
             remaining: u.get("remaining").and_then(|t| t.as_u64()).unwrap_or(0),
         }),
+        additional_metadata_args: AdditionalMetadataArgs {
```
austbot, line 284: a-lot of these fields are already there.

austbot (changes requested):
hey so name, symbol, mutability are all duplicated  whats the rationale for duplicating?

PR #28, opened 2022-11-13
candy guard ingester and api changes

stranzhay:
finishing off candy guard ingester,api code 

Screen Shot 2022-11-14 at 4 59 21 PM


stranzhay:
there is an issue for how the db is reading mint authority from cm table and matching it to plerkle pubkey coming into ingester. the fk linking of a candy guard id to a cm table is working fine as the screenshot above shows.

staccDOTsol:
It looks like you're working on finishing off the Candy Guard ingester and API code. Can you provide more context or specific questions about what you're trying to accomplish or what issues you're facing?

PR #42, opened 2023-03-18
fix: materialize cl_items to decrease locking

helius-wedtm:
Index updates will lock the table, as the size increases index update latency increases O(n) with table size.

Using a materialized view will only lock the view, not the original table, allowing ingest to continue.

This comes at the expense of the view becoming stale, and requiring `REFRESH MATERIALIZED VIEW cl_items_view;` to be executed on a business-level cadence.

Thank you for taking the time to make an enhancement or bug fix to the digital asset validator rpc repo.
To ensure a quality PR experience this call may be recorded. Just Kidding, but we do expect a few things.

1. The Name of the PR will show up in the changelog so make it a good one, we will rename PRs or reject based on the name.
2. Please make the PR as small as possible to achieve the bugfix or feature. Big prs often are scary and hard to review.
3. Be kind to your reviewers 🤓
4. Add a good description, so we can see what the PR is all about without investing the time in the code review. We will often review code at a certain time in the day and having a list if important prs to review helps us move that along.

### DIFF
--- a/init.sql
+++ b/init.sql
@@ -20,18 +20,23 @@ CREATE TABLE cl_items
     hash     BYTEA  NOT NULL
 );
 -- @@@@@@
--- Index All the things space is cheap
-CREATE INDEX cl_items_tree_idx on cl_items (tree);
+-- Index All the things space is cheap, but index rebuilds are not
+-- Let's materialize cl_items into a view so the indexes don't lock
+-- the cl_items table and prevent ingest
+CREATE MATERIALIZED VIEW cl_items_view AS
+SELECT * FROM my_table;
+
+CREATE INDEX cl_items_tree_idx on cl_items_view (tree);
 -- @@@@@@
-CREATE INDEX cl_items_hash_idx on cl_items (hash);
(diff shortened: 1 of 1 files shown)


linuskendall:
Question - wouldn't refreshing a view meaning complete recomputation of all the indexes? As opposed to the current state where the indexes are only updated based on the updates to the source tables? AFAIR refresh of a materialized view completely recreates it.

staccDOTsol:
It seems like you're proposing a solution to decrease locking on the `cl_items` table by using a materialized view. This is a good approach to reduce the locking time on the original table, but as you mentioned, it comes at the expense of the view becoming stale. 

To mitigate this, you can schedule a job to refresh the materialized view on a regular basis, based on your business needs. This will ensure that the view is up-to-date and the data is not stale. 

Overall, this is a good solution to reduce locking time on the original table and improve ingest performance.

helius-wedtm:
Sorry folks, this was meant as a WIP PR that I forgot to label as a draft!

Issue #51, opened 2023-04-05
GetAssetsByCreator is returning random assets for random pubkeys.


ethyi:
Inputting a random pubkey into getAssetsByCreator will return a random set of data. Data just so happens to be alphanumerically sorted, perhaps for the entire DB.
```
{
    "jsonrpc": "2.0",
    "id": "0",
    "method": "get_assets_by_creator",
    "params": {
        "page": 1,
        "creatorVerified": true,
        "creatorAddress": "1234abcd8htfC7YEXKNEfkpuogUUQDDTLgjGWxxv48a",
        "limit": 1000
    }
}
```

brianlong:
🤦🏻

NicolasPennie:
FYI, the proper query would have "verifiedOnly" instead of "creatorVerified". 

When "verifiedOnly" = true, it works as intended.
When "verifiedOnly" = false, every single non-verified NFT is returned.

NicolasPennie:
Fixed via https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/issues/51

Unfortunately there is still a problem with regular NFTs: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/issues/53

PR #52, opened 2023-04-06
Fix GetAssetsByCreator query

NicolasPennie:
## Summary
* Fixes https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/issues/51
* The SQL was querying for "creator = X OR supply > 0" instead of "creator = X AND supply >0"
* I removed the multiple creators from the query since we only ever feed a single creator to the method. Simpler code = less likely for bugs.

## Testing
* Tested e2e in the Helius dev environment with mainnet data. Ensured the GetassetsByCreator returns the expected for both cases of `only_verified=true/false`.

### DIFF
--- a/das_api/Cargo.lock
+++ b/das_api/Cargo.lock
@@ -992,7 +992,7 @@ dependencies = [
 
 [[package]]
 name = "das_api"
-version = "0.6.8"
+version = "0.7.1"
 dependencies = [
  "async-trait",
  "bs58 0.4.0",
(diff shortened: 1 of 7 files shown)


linuskendall (approved):
Looks good to me.  Do we need to update some documentation for the change of get_asset_by_creator -> get_asset_by_creators

NicolasPennie:
> Looks good to me. Do we need to update some documentation for the change of get_asset_by_creator -> get_asset_by_creators

The public API does not take an array. It was taking a fixed address and mapping it to an array of length 1 each time. So to simplify the code I changed it so that its not an array.

linuskendall:
I went ahead and merged this, hope you dont mind @NicolasPennie

Issue #50, opened 2023-04-05
Offchain Indexers unreliably processing data


ethyi:
The offchain indexer is failing to reliably process offchain json uris. 
This is making get_asset calls return NFTs without its offchain data populated, the most important being the 'files' field which usually holds the uri to the NFT's image.

## Context
NFT metadata gets first indexed when the metadata account is processed via acc geyser. Note that this is different from the mint account.
The ingester initially populates the DB with the onchain metadata, a reference to the offchain uri, and puts a placeholder "processing" in the metadata field until the offchain indexer populates it. 
The offchain indexer is a background task that is asynchronously invoked by the ingester. When a new task is submitted it is added to the "tasks" table, and each row contains data on the id that called it and its status. 

The task makes a get request to the uri with a timeout of 3 seconds. If at any point the request returns a "reqwest::Error", the row in the tasks table will detail that a failure happened along with a short message identifying the error. Currently all reqwest::Error are converted to "Error downloading batch files" which is likely what most will receive.

- Each NFT will create its own background task, regardless if many have the same json uri
- The bgtask runner will purge tasks older than 1 hr
- If the task is already present in the database, a new task job will not be submitted even if the metadata account is reprocessed.
  - Owning an FFF is a great way to test this since you can dynamically change the offchain data. 
- The timeout for offchain calls is 3 seconds.
- There are a maximum of 3 attempts before the offchain indexer completely drops the data. 
- The bgtask runner will rerun tasks for all rows in asset_data that have their metadata as "processing"
- (*on helius) there are more offchain failures than there are successes

## Examples
3LBkptXr6N9y3S2eSRTxCega91qHd4Fd6hWyNmBdCzRu
## Solutions
- Priority background task runner with more flexibility. Addresses that see lots of activity will use this instead
- Increase attempts, and increase timeout if possible.

linuskendall:
I took a bit of look at this and a few ideas from my side:

1) We should add a configurable timeout and a configurable number of retries, probably also MAX_TASK_BATCH_SIZE should be configurable too so we could tweak these values accordingly. In this way we can experiment with different values for these and track results (see 2)

2) We should capture the metrics for error rates and compare, right now it looks like it's quite different across different envs. We have one env where it's about 30% at the moment. Is the bgtask_succes/bgtask_error sufficiently finegrained metric or is there another one we should look at ?

[...]

PR #56, opened 2023-04-17
Configurable bgtask runner

linuskendall:
Add configuration to the task runners, this helps address problems outlined in #50.

Adds support for a config file in addition to the environment variables. Environment variables overrides the config file values.

### DIFF
--- a/nft_ingester/src/tasks/mod.rs
+++ b/nft_ingester/src/tasks/mod.rs
@@ -377,12 +377,12 @@ impl TaskManager {
                     Ok(res) => {
                         debug!("deleted {} tasks entries", res.rows_affected);
                         metric! {
-                            statsd_count!("ingester.bgtask.purged_tasks", i64::try_from(res.rows_affected).unwrap_or(1), "type" => &task_name);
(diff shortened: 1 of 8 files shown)


nft_ingester/src/tasks/mod.rs
```diff
@@ -233,8 +276,8 @@ impl TaskManager {
         })
     }
 
-    pub async fn purge_old_tasks(conn: &DatabaseConnection) -> Result<DeleteResult, IngesterError> {
-        let cod = Expr::cust("NOW() - created_at::timestamp > interval '60 minute'"); //TOdo parametrize
+    pub async fn purge_old_tasks(conn: &DatabaseConnection, task_max_age: time::Duration) -> Result<DeleteResult, IngesterError> {
+        let cod = Expr::cust_with_values("NOW() - created_at::timestamp > interval '? seconds'", [task_max_age.as_secs()]);
```
linuskendall, line 280: This is broken, I think it doesn't actually do replacement.

austbot (approved):

PR #58, opened 2023-04-18
Fix ordering of asset metadata processing

linuskendall:
Ensure asset metadata is processed by attempts, so that we don't keep retrying the same asset over and over.


nft_ingester/src/tasks/mod.rs
```diff
@@ -158,7 +158,8 @@ impl TaskManager {
                             .less_or_equal(Expr::col(tasks::Column::MaxAttempts)),
                     ),
             )
-            .order_by_desc(tasks::Column::CreatedAt)
+            .order_by(tasks::Column::Attempts, Order::Asc)
+            .order_by(tasks::Column::CreatedAt, Order::Desc)
```
NicolasPennie, line 162: Shouldn't this order createdAt by descending? We want to treat the tasks like a queue.

austbot: That was the intent to process newer as higher priority

linuskendall: I belive these comments are addressed? It processes newer records as they come in but will downgrade older ones.

danenbm (approved):
This LGTM but lets not merge until we get the creator/data hash fix merged.

PR #60, opened 2023-04-19
Account forwarder

ethyi:
Acc forwarder is a testing tool for simulating the account input from geyser.
Useful for debugging regular NFT indexing.

There are multiple modes offered

Single: for sending 1 account
Mint: for indexing the mint, its metadata, and its owner
for sending multiple accounts


ethyi:
Unfortunately don't have write access

PR #68, opened 2023-05-08
fix(das-api): fix getAssetProof

NicolasPennie:
Empty proofs were being provided once we reached the 50% of the tree size.


linuskendall (approved):
LGTM

PR #77, opened 2023-05-24
Enable out-of-order transaction processing for `asset` table

danenbm:
### Notes
* Update owner and delegate in asset table when collection or creator verification occurs.
* Modify program transformers to enable out-of-order transaction processing by adding the following changes:
  * Upsert `leaf` info based on `seq` OR `was_decompressed` flag.
  * Upsert `owner` and `delegate` based on `owner_delegate_seq`.
  * Upsert "compression status" fields (`compressed`, `compressible`, `supply`, and `supply_mint`) based on `was_decompressed` flag.

### Database changes
* Add `owner_delegate_seq` and `was_decompressed` columns to asset table, and `PROGRAMMABLE_NFT` to `specification_asset_class` type.
  * Regenerate Sea ORM objects.
  * Needed to temporarily comment out a performance improvement migration that removed foreign keys so that SeaORM CLI would regenerate, since it builds relations based on those constraints.
  * Needed to add a migration to add `PROGRAMMABLE_NFT` to `specification_asset_class` since it is needed by the API code but missing from the SQL code.

### Other misc. changes
* Fix docker preparation script to build solana-program-library.
* Also remove unused 'filling' variable.
* Rustfmt/clippy fixes as I change files.

### Testing
  1. Made sure it builds and runs in Docker.
  2.  Read API specific test: Ran a sequence of `CreateTree`/`Mint`, `Transfer`, and `Burn` in reverse order.  While running in reverse order I observed that could not get info for the asset via `getAsset` until `Mint` was indexed.
  3. Used transaction forwarder to send out sequences of transactions in correct and reverse order, and in each case observe `cl_items` and `asset` table in database.  Covered every indexed instruction (note `MintToCollection` indexes as `Mint`).  Did this process on both main branch and this PR branch and compared all results.

  | Transaction sequence | Forwards on main | Backwards on main | Forwards on PR branch | Backwards on PR branch |
  | ----------- | ----------- | ----------- | ----------- | ----------- |
  | `CreateTree`/`Mint`, `Transfer`, and `Burn` | OK | `cl_items` OK, `asset` incorrect due to only indexing `Mint` | OK | OK |
  | `CreateTree`/`Mint`, `Redeem`, `CancelRedeem`, second `Redeem`, and `Decompress` | `cl_items` and`asset` incorrect due to `Redeem` not indexing | `cl_items` incorrect due to `Redeem` not indexing, `asset` incorrect due to only indexing `Mint` | OK | OK |
  | `CreateTree`/`Mint`, `Transfer`, second `Transfer` | OK | `cl_items` OK, `asset` incorrect due to only indexing `Mint` | OK | OK |
  | `CreateTree`/`Mint`, `Delegate`, `Transfer` | OK | `cl_items` OK, `asset` incorrect due to only indexing `Mint` | OK | OK |
  | `CreateTree`/`Mint`, `VerifyCreator` | OK | `cl_items` OK, `asset` incorrect due to only indexing `Mint`, `asset_creators` incorrect due to missing `VerifyCreator` | OK | OK, fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/87 ~~`cl_items` and `asset` OK, `asset_creators` incorrect due to missing `VerifyCreator`~~ |
  | `CreateTree`/`Mint`, `VerifyCollection` | `cl_items` and `asset` OK, `asset_grouping` incorrect due to missing `VerifyCollection` | `cl_items` OK, `asset` incorrect due to only indexing `Mint`, `asset_grouping` incorrect due to missing `VerifyCollection` | OK, fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/90. `SetAndVerifyCollection` still needs Blockbuster fix to index. ~~cl_items and asset OK, asset_grouping incorrect due to missing VerifyCollection~~ | OK, fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/90. `SetAndVerifyCollection` still needs Blockbuster fix to index. ~~cl_items and asset OK, asset_grouping incorrect due to missing VerifyCollection~~|


NicolasPennie:
What's the motivation for this PR? Was anyone observing indexing issues due to out of order transactions?

danenbm:
> What's the motivation for this PR? Was anyone observing indexing issues due to out of order transactions?

What I heard from @austbot is that both Helius and Triton had observed issues with out of order transactions and as a result, had to essentially stop running the backfiller and use a different method for in-order filling.  I also talked to @austbot and @linuskendall about this and it does in fact look like this reference code could not handle out of order transactions in general.

The `cl_table` looks like it can handler out of order transactions, but the `asset` table does not.  The reason the `cl_table` can handle out of order transactions is because all instructions except for `burn` and `decompress` update the same leaf schema fields, and use the tree sequence number to prevent stale data.

For example, if a `transfer` was ingested followed by a `mint` (out of order), the change log will be inserted into `cl_items` for the `transfer`, and then when the `mint` is ingested, it won't update the fields in the table with stale data (owner, delegate, etc.) because the sequence number for the `mint` will be a lower number than the sequence number for the `transfer` that was already ingested.

However, using this same example but considering the `asset` table, what happens is when the `transfer` comes in first, is the asset is not found in the table, because the `mint` did not get indexed yet.  So the `update` fails and logs a message and the indexer continues.  Later when the `mint` instruction is indexed, it inserts the row for the asset into the asset table.  Now the `asset` table contains the asset with a stale owner, delegate, and leaf hash.

NicolasPennie:
> > What's the motivation for this PR? Was anyone observing indexing issues due to out of order transactions?
> 
> @NicolasPennie what I heard from @austbot is that both you (Helius) and Triton had observed issues with out of order transactions and as a result, had to essentially stop running the backfiller and use a different method for in-order filling. I also talked to @austbot and @linuskendall about this and it does in fact look like this reference code could not handle out of order transactions in general.
> 
> The `cl_table` looks like it can handler out of order transactions, but the `asset` table does not. The reason the `cl_table` can handle out of order transactions is because all instructions except for `burn` and `decompress` update the same leaf schema fields, and use the tree sequence number to prevent stale data.
> 
> For example, if a `transfer` was ingested followed by a `mint` (out of order), the change log will be inserted into `cl_items` for the `transfer`, and then when the `mint` is ingested, it won't update the fields in the table with stale data (owner, delegate, etc.) because the sequence number for the `mint` will be a lower number than the sequence number for the `transfer` that was already ingested.
> 
> However, using this same example but considering the `asset` table, what happens is when the `transfer` comes in first, is the asset is not found in the table, because the `mint` did not get indexed yet. So the `update` fails and logs a message and the indexer continues. Later when the `mint` instruction is indexed, it inserts the row for the asset into the asset table. Now the `asset` table contains the asset with a stale owner, delegate, and leaf hash.
> 
> This PR addresses this issue with the `asset` table, allowing for out-of-order Bubblegum transactions to be processed in general. I am still looking at the edge cases and how this works for non-compressed NFTs.

Great, thanks for the context! That was really helpful.

austbot (approved):
reviewed on a call with md

danenbm:
> this is LGTM @danenbm

Thank you.  Still testing this and working through some issues with upserting parts of an asset row due to NULL constraints on some of the fields.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -130,42 +134,83 @@ where
...
 
-                // Do not attempt to modify any existing values:
-                // `ON CONFLICT ('id') DO NOTHING`.
-                let query = asset::Entity::insert(model)
+                // Upsert asset table base info.
+                let query = asset::Entity::insert(asset_model)
                     .on_conflict(
                         OnConflict::columns([asset::Column::Id])
-                            .do_nothing()
+                            .update_columns([
```
linuskendall, line 161: question - if it's possible in the contract to see two mint_v1 for the same asset at different slots (?) then this could overwrite newer data with older data.

danenbm: It is not possible to have two `mint_v1` for the same `asset_id`, because in the contract, the `asset_id` is a PDA where one of the seeds is the `num_minted` for the specific tree.  Upon successful mint, `num_minted` is incremented.  It is monotonically increasing so we should never end up with the same `num_minted` for a given tree.

digital_asset_types/src/dao/generated/sea_orm_active_enums.rs
```diff
@@ -116,10 +106,20 @@ pub enum SpecificationAssetClass {
     Print,
     #[sea_orm(string_value = "PRINTABLE_NFT")]
     PrintableNft,
+    #[sea_orm(string_value = "PROGRAMMABLE_NFT")]
+    ProgrammableNft,
     #[sea_orm(string_value = "TRANSFER_RESTRICTED_NFT")]
```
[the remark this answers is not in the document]
danenbm: Why is that, because we won't ever use it?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -114,42 +111,127 @@ where
...
+        .build(DbBackend::Postgres);
+
+    // If we are indexing decompression we will update the leaf regardless of if we have previously
+    // indexed decompression and regardless of seq.
+    if !was_decompressed {
+        query.sql = format!(
+            "{} WHERE (NOT asset.was_decompressed) AND (excluded.seq > asset.seq OR asset.seq IS NULL)",
+            query.sql
+        );
     }
```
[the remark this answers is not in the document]
danenbm: Ok nevermind about this issue above, I fixed it with this commit: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/77/commits/f7a51430cef898f76cebd3f3aaaa74a9eff3c903

nft_ingester/src/program_transformers/bubblegum/cancel_redeem.rs
```diff
@@ -19,29 +21,43 @@ where
...
-                    leaf: Set(Some(le.leaf_hash.to_vec())),
-                    delegate: Set(delegate),
-                    owner: Set(Some(owner_bytes)),
-                    seq: Set(seq as i64), // gummyroll seq
-                    ..Default::default()
-                };
-                update_asset(txn, id_bytes, Some(seq), asset_to_update).await
+
+                // Partial update of asset table with just leaf.
+                upsert_asset_with_leaf_info(
```
NicolasPennie, line 41: Have you factored in that these are now separate DB writes and on failure, it will only be partially complete?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -114,42 +111,157 @@ where
     //TODO -> set maximum size of path and break into multiple statements
 }
 
-pub async fn update_asset<T>(
+pub async fn upsert_asset_with_leaf_info<T>(
```
NicolasPennie, line 114: Why do we need to split the upserts between leaf / owner+delgate? Can't we handle it the same via a single upsert?

To attempt to answer my own question:

I assume the problem is that the seq value for a newer txn will be higher than the mint seq, meaning that some of the mint info will never get added to the asset table. So we need to break apart our seq value to distinguish between different attribute updates.

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -114,42 +111,157 @@ where
...
+                .update_columns([
+                    asset::Column::Owner,
+                    asset::Column::Delegate,
+                    asset::Column::OwnerDelegateSeq,
+                ])
+                .to_owned(),
+        )
+        .build(DbBackend::Postgres);
+    query.sql = format!(
+        "{} WHERE excluded.owner_delegate_seq > asset.owner_delegate_seq OR asset.owner_delegate_seq IS NULL",
```
NicolasPennie, line 185: delegate_seq is only null at first right? never later, right?

nft_ingester/src/program_transformers/bubblegum/transfer.rs
```diff
@@ -21,29 +22,43 @@ where
...
+                upsert_asset_with_owner_and_delegate_info(
+                    txn,
+                    id_bytes.to_vec(),
+                    owner_bytes,
+                    delegate,
+                    seq as i64,
+                )
+                .await?;
+
+                upsert_asset_with_seq(txn, id_bytes.to_vec(), seq as i64).await
```
NicolasPennie, line 61: Is this necessary? I think we can just update this within one of the earlier methods. We try to reduce to number of DB calls we need to make in favour of performance. (applies for all cNFT indexing methods).

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -130,42 +134,85 @@ where
...
+                                asset::Column::SlotUpdated,
+                                asset::Column::DataHash,
+                                asset::Column::CreatorHash,
+                            ])
                             .to_owned(),
                     )
                     .build(DbBackend::Postgres);
                 txn.execute(query).await?;
 
+                // Partial update of asset table with just compression info elements.
```
NicolasPennie, line 182: Breaking this apart into multiple upserts will put more load on the DB. Did you an analysis on the DB load changes? Did you test this under high load (e.g. 1kTPS+)?

nft_ingester/src/program_transformers/bubblegum/burn.rs
```diff
@@ -26,27 +33,32 @@ where
...
         debug!("Indexing burn for asset id: {:?}", asset_id);
-        let id_bytes = asset_id.to_bytes().to_vec();
-        let asset_to_update = asset::ActiveModel {
-            id: Unchanged(id_bytes.clone()),
+        let id_bytes = asset_id.to_bytes();
+
+        let asset_model = asset::ActiveModel {
+            id: Set(id_bytes.to_vec()),
             burnt: Set(true),
-            seq: Set(seq as i64),
```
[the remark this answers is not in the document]
danenbm: If a burn transaction comes in out of order, it needs to set the `burnt` flag (because nobody else will as it's not a leaf schema thing), but we don't want it to overwrite `seq` with a potentially older value.  Thus `seq` is now updated later by `upsert_asset_with_seq`, which will only update `seq` if `excluded.seq > asset.seq`.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Correct, it is only null when when the asset row is created by some other upsert that does not put a value there. And yes it should never go back to null after its been populated.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Yes exactly right.  In order to handle out-of-order we need to break apart `seq` value for each "block" of data that can be updated by different instructions.

To use your example, the `seq` value for a `transfer` will be higher than the `seq` value when the asset was minted.  So if it was all upserted but controlled by the same `seq`, and if we saw the `transfer` come through first, then we would get the `leaf`, `owner`, and `delegate` from the `transfer` event, but we would never get the `specification_asset_class` which is something only set during `mint`.

That is why `mint` first does all the mint-only data, then `leaf`, `owner`, and `delegate` are separate.

Then to address why `leaf` is separate than `owner`+`delegate`,  it is because `redeem` and `decompress` update the `leaf` but do not update the `owner` and `delegate`.  So for example if `redeem` was indexed before a `transfer` that actually preceeded it on chain, and if they were all controlled by the same `seq`, then the `owner` and `delegate` would not be updated to their correct values.  `owner` and `delegate` can be grouped together because they are always updated as a pair.

This same phenomenon is true with a block of data I am calling "compression info" which is updated by both `mint` and `decompress`.  And this phenomenon is even true with the base `seq` number itself, if we want to keep that accurate in the asset table.

The general form of this whole PR's approach, which i think you understand already, is that:
1. Data updated by a single instruction can be indexed in any order at any time.
2. Data updated by multiple instructions need to be protected by a sequence number for that specific data, and all instructions that update that particular data must update that data's sequence number.

README.md
```diff
@@ -38,6 +38,10 @@ Then with a local `DATABASE_URL` var exported like this `export DATABASE_URL=pos
 
 If you need to install `sea-orm-cli` run `cargo install sea-orm-cli`.
 
+Note: The current SeaORM types were generated using version 0.9.3 so unless you want to upgrade you can install using `cargo install sea-orm-cli --version 0.9.3`.
+
+Also note: The migration `m20230224_093722_performance_improvements` needs to be commented out of the migration lib.rs in order for the Sea ORM `Relations` to generate correctly.
```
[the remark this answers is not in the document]
danenbm: According to @austbot we need this performance improvement, but the problem is the autogeneration relies on the foreign key constraints to work.  So I think I'm with you that this could be removed, but I didn't want to do it as part of this PR.  I think it would be better to do it as a follow on PR just focused on the effects of those changes.

nft_ingester/src/program_transformers/bubblegum/cancel_redeem.rs
danenbm: Yes I have thought about it seems like it is is OK, at least as OK as the current code with how it deals with DB write failures.

What I mean is that in the existing code, if (for example) a database write fails on `transfer`, then you just get a metric or log and the asset record for  `leaf`, `owner`, `delegate` and `seq` would be unchanged.  The `transfer` transaction would need to be rerun to fix it (which isn't 100% reliable given shared `seq`), or a later transaction that updates those fields.  So in the existing code when this happens the `asset` row is overall just partially correct.

With this PR's change, you could have a state where 'leaf` had been updated correctly but `owner`, `delegate` and `seq` may be out of date.  The `transfer` transaction or a later transaction would still need to be rerun to fix them, similar as before.  So again the overall state is one of partial correctness.  Just different partially correct fields than before.

Neither case auto-corrects for the failed write as far as I know and that would be a more extensive change to the backfilling process.

nft_ingester/src/program_transformers/bubblegum/transfer.rs
danenbm: In a previous iteration of this PR I used the existing `seq` to protect `leaf` updates, and the new `owner_delegate_seq` to protect `owner` and `delegate` fields. Because of this, an effect of my PR changes is that when the leaf was set to `None` during decompression, so was the `seq`, and then for decompressed assets, a Read API call to `getAsset` would return `null` for the `seq`. 

Maybe that was OK because people don't need the `seq` when the asset has been removed from the merkle tree. But the merkle tree is still listed and wanted it to match previous behavior.

Maybe I can change it back if we want to remove an extra update that is only for the `seq`.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: So overall for this PR I just need to determine final strategy to complete it.  Here are the options I see right now based on these comments:

1. Just test it under load with the multiple upserts as-is, and if its working fine under high load, leave it as-is.
2. Modify the code to have one upsert followed by 1-2 updates for each program operation.  I initially chose all upserts because it was simpler code and less prone to issue if someone changes order of things, but updates will be less load than upserts. 
3. Modify the code to just use the `seq` for `leaf` and `owner_delegate_seq` for `owner`+`delegate`.  This removes one of the upserts/updates from each operation.

NicolasPennie: I would test it under load beforehand and then see how it holds up under load afterwards. Especially curious about memory stress. We have been running into issues with a 256GB RAM DB when we're under higher loads.

I'll be carving out some time to do a deeper analysis to figure out the conditions that specifically cause the pain. I believe its mostly due to the high number of indexes and updating all of them quickly is expensive on memory. Especially as the cl_items table grows quite large.

nft_ingester/src/program_transformers/bubblegum/transfer.rs
NicolasPennie: > a Read API call to getAsset would return null for the seq.
This would be acceptable if the asset can never be re-compresssed. But I think you plan on adding that, right?

If that's the case, then it makes sense to keep the seq value available after a decompression.

nft_ingester/src/program_transformers/bubblegum/cancel_redeem.rs
NicolasPennie: That was also my take. Works for me.

nft_ingester/src/program_transformers/bubblegum/burn.rs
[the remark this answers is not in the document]
NicolasPennie: Hmm gotcha.

Can there be any events for a cNFT after its burnt?

README.md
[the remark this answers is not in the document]
NicolasPennie: Ah gotcha, this rings a bell from when we added a migration ourselves and it ruined all of our generated models.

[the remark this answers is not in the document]
NicolasPennie: for what its worth, we don't need the relations to exist on the models to power those queries. it's just a nice to have. we can just restructure the SeaORM query and it will work fine.

nft_ingester/src/program_transformers/bubblegum/burn.rs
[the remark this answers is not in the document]
danenbm: No there cannot be any events for a cNFT after its burnt.

1. It is intended that burn destroys the NFT, which it does by replacing that leaf with zeros, leaving an empty hole in the tree.
2. If someone tried to do something with the leaf after burning it, any of those operations fail because the "previous leaf" is determined in other instructions by id, owner, data_hash, etc. which would not match the leaf of all zeros.

nft_ingester/src/program_transformers/bubblegum/transfer.rs
danenbm: Yeah I think long term we plan to add re-compression.  We have the placeholder for re-compression but not fully designed yet.  Maybe it would insert back into the same tree like `cancel_redeem` does but instead of using a voucher it would burn the token-metadata based NFT.  But I could also see it just doing a new append to the tree via `mint_v1`.

Either way when it got added back into the tree, I think the asset would get a new `seq` that is probably a much higher number that what was previously stored there when it was decompressed, as many transfers and other operations probably happened on the tree while this NFT was decompressed.

I guess what I'm thinking at the moment is that I'm not sure of the use for the `seq` after decompression since you cannot use it for re-compression, right?  Or am I misunderstanding something?

README.md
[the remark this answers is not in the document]
danenbm: Ok that sounds good too.  I haven't looked into that yet.  Maybe I can do it as a follow on PR unless its really easy to just do it now.

nft_ingester/src/program_transformers/bubblegum/transfer.rs
danenbm: For reference the previous version of my PR prior to [this commit](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/77/commits/f7a51430cef898f76cebd3f3aaaa74a9eff3c903) used `seq` to protect leaf and `owner_delegate_seq` to protect `owner`+`delegate` and had one less upsert overall.  May be worth revisiting that approach to save on an update/upsert.

NicolasPennie: > I guess what I'm thinking at the moment is that I'm not sure of the use for the seq after decompression since you cannot use it for re-compression, right?

If recompressing will add that NFT back into the tree via the same leaf, then we need to keep the seq number around. Otherwise you can have out of order bugs with out of order decompress + recompress transactions. And we need to be able to always handle any order to be safe, even if its unlikely to happen.

danenbm: I think you uncovered something that will be an issue for `decompress` > `recompress` with the current design both of this PR and with the system in general.  The fact that decompress doesn't have any association with the tree, means there is no change log event and thus no `seq` associated with the `decompress`.

For this PR, the way I found to deal with `decompress` being out of order with others was to add a boolean `was_decompressed` flag, which essentially acts like a 1-bit sequence number, which is a one way door assuming no events after the decompression.

nft_ingester/src/program_transformers/bubblegum/redeem.rs
```diff
@@ -19,34 +17,33 @@ pub async fn redeem<'c, T>(
...
+        let leaf_index = cl.index;
+        let (asset_id, _) = Pubkey::find_program_address(
+            &[
+                "asset".as_bytes(),
+                cl.id.as_ref(),
+                u32_to_u8_array(leaf_index).as_ref(),
+            ],
+            &mpl_bubblegum::ID,
+        );
+        debug!("Indexing burn for asset id: {:?}", asset_id);
```
NicolasPennie, line 31:
```suggestion
        debug!("Indexing redeem for asset id: {:?}", asset_id);
```

nft_ingester/src/program_transformers/bubblegum/transfer.rs
austbot: @NicolasPennie @danenbm we must keep the seq in the db if the asset is recompressed. 
there are other things though, since you cant close the mint acccount you cannot fully recompress. so you can use that as a handle  some how.

danenbm: I still don't think recompression can work out of order, because decompression is not sequenced by the tree.  Am I missing something?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -114,42 +111,157 @@ where
...
     }
+
+    txn.execute(query)
+        .await
+        .map_err(|db_err| IngesterError::StorageWriteError(db_err.to_string()))?;
+
+    Ok(())
+}
+
+pub async fn upsert_asset_with_owner_and_delegate_info<T>(
```
nikhil-helius, line 155: Hi! Any reason why `upsert_asset_with_leaf_info` and `upsert_asset_with_owner_and_delegate_info` happens as two separate queries on the `asset` table instead of a single query?

danenbm: Yes there is a reason.

First, the general form of this PR's strategy across all the indexed data is that to maintain correctness when instructions are processed out of order:
* Any "block" of data updated by a single instruction can be indexed in any order at any time.
* Any "block" of data updated by multiple instructions need to be protected by a sequence number for that specific data, and all instructions that update that particular data must update that data's sequence number.

So specifically the reason `leaf` is separate from `owner`+`delegate` is because `redeem` and `decompress` update the `leaf` but do not update the `owner`+`delegate`.

So for example if `redeem` was indexed before a `transfer` that actually preceeded it on chain, and if they were all controlled by the same `seq`, then the `owner`+`delegate` would not be updated to their correct values.

 `owner`+`delegate` can be grouped together with one sequence number for both of them because they are always updated as a pair whenever there is a leaf schema event.

nikhil-helius: gotcha, thanks. I understand the need to separate it for `redeem` and `decompress` but for all the cases where `upsert_asset_with_owner_and_delegate_info` is called, we update the leaf as well. We should still have separate sequence numbers for `leaf` and `owner+delegate` but it those cases where we're updating both `leaf` and `owner, delegate`, couldn't they be a part of a single tx? Wouldn't it faster and lesser load during writes?

danenbm: I had them all as separate upserts for simplicity in the code, but if I can combine them for better performance I yes I will do that.

So just so I understand, do you mean creating two SQL statements but they are included in one single executed transaction?  Something like "update leaf where seq > leaf seq", and "update owner/delegate where seq > owner_delegate_seq", both added to the same tx, probably separated by semicolon when it is actually turned into SQL?

nikhil-helius: Yes, that's what I meant. I don't know if its possible to have a single query since each of them separate conditions on them. Let me also check if that's possible but yes, would be good if its a single transaction.

danenbm: Was talking with the team and they suggested that we merge this PR, https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/87, and https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/90 which all have this same `upsert` strategy, and then after they are mainlined we can focus on optimizations.

nikhil-helius: sounds good. Thanks!

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: Tested by Triton on their non-prod environment.  Also Helius doing some testing.  Marking as resolved.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Ok resolving this for now will focus on optimization after we nail correctness across all instructions!

PR #79, opened 2023-06-06
feat(das-api): support searching assets by `jsonUri` parameter

tsmbl:
Support optional filtering by `jsonUri` in `searchAssets` API method

### DIFF
--- a/das_api/src/api/mod.rs
+++ b/das_api/src/api/mod.rs
@@ -74,12 +74,12 @@ pub struct SearchAssets {
     pub royalty_target: Option<String>,
     pub royalty_amount: Option<u32>,
     pub burnt: Option<bool>,
-    pub json_uri: Option<String>,
     pub sort_by: Option<AssetSorting>,
     pub limit: Option<u32>,
     pub page: Option<u32>,
     pub before: Option<String>,
     pub after: Option<String>,
+    pub json_uri: Option<String>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
--- a/das_api/src/api/mod.rs
+++ b/das_api/src/api/mod.rs
@@ -74,6 +74,7 @@ pub struct SearchAssets {
     pub royalty_target: Option<String>,
     pub royalty_amount: Option<u32>,
     pub burnt: Option<bool>,
+    pub json_uri: Option<String>,
     pub sort_by: Option<AssetSorting>,
     pub limit: Option<u32>,
     pub page: Option<u32>,
@@ -83,7 +84,6 @@ pub struct SearchAssets {
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "camelCase")]
-
 pub struct GetAssetsByAuthority {
     pub authority_address: String,
     pub sort_by: Option<AssetSorting>,
--- a/das_api/src/api/mod.rs
+++ b/das_api/src/api/mod.rs
@@ -79,6 +79,7 @@ pub struct SearchAssets {
     pub page: Option<u32>,
     pub before: Option<String>,
     pub after: Option<String>,
+    #[serde(default)]
     pub json_uri: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -172,27 +171,40 @@ where
 
 pub async fn get_related_for_assets(
     conn: &impl ConnectionTrait,
-    assets: Vec<(asset::Model, Option<asset_data::Model>)>,
+    assets: Vec<asset::Model>,
```
austbot, line 174: whats going on here?

austbot (commented):
Hey thanks for this contribution, ill let @danenbm have the final say as im a thirdparty contributor now. But it looks to me like you have the start of something great. Just had one question.

digital_asset_types/src/dao/scopes/asset.rs
tsmbl: Long story short: previously the join between `asset` and `asset_data` was done outside of `get_related_for_assets` fn, but I've moved it inside `get_related_for_assets` body, therefore function signature is changed, the reasoning is below

Before my changes, there was a single inclusion of `asset_data` table in resulting sql query: `asset_data` was included in query because of joining to `asset` via `find_also_related(asset_data::Entity)` statement  in `get_assets_by_condition` function,  just before `get_related_for_assets` was called. 

Since my PR involved filtering by `asset_data.json_uri`, it lead to the inclusion of `asset_data` in filter query, that made impossible to use `find_also_related(asset_data::Entity)` statement  in `get_assets_by_condition` function, because it produced the error `Database Error Query Error: error returned from database: table name \"asset_data\" specified more than once"`.

To negotiate this, I've decided to remove the 2nd inclusion of `asset_data` in sql query by removing `find_also_related(asset_data::Entity)` in favour of explicit joining of `asset_data` to `asset` inside `get_related_for_assets()` function, therefore `get_related_for_assets` function signature is changed

@austbot, thoughts?

tsmbl:
@austbot thank you for the prompt review! I hope my explanation to your question is reasonable, please let me know if you have any further questions

tsmbl:
> I see were you able to test all the other methods to ensure it works for them? 

yes, I've manually tested other api methods that utilize the changed codebase

das_api/src/api/mod.rs
```diff
@@ -74,6 +74,7 @@ pub struct SearchAssets {
     pub royalty_target: Option<String>,
     pub royalty_amount: Option<u32>,
     pub burnt: Option<bool>,
+    pub json_uri: Option<String>,
```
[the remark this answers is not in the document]
tsmbl: Nice catch, will fix

tsmbl:
@austbot btw, curious about your thoughts on adding idx for `json_uri` attribute

Upon reconsideration, I decided against adding an index for the `json_uri` attribute due to the following reasons:

1. I've noticed that commits have been made to remove indexes and constraints for better asset indexing performance
2. Non-indexed attributes are still searchable, indicating that the system can handle search queries efficiently without indexes on all fields
3. Since the `grouping` is already indexed and we will also filter by it in our use case, adding an extra index for `json_uri` may not be necessary

danenbm (commented):
Reading through the comments, taking a look at this PR tomorrow.

das_api/src/api/mod.rs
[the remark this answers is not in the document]
tsmbl: imo, ideally array params should be disabled, since it’s error prone and causes compatibility problems, so prob we need api v2 when we want to add more param

linuskendall (approved):
At a look it looks good to me - I think this was tested for backwards compatibility because that's really the main concern.

Re indexing I would not index the json_uri, but we can do some individual testing of query speeds after this support is merged and make that a specific optimisation?

digital_asset_types/src/dao/scopes/asset.rs
danenbm: This makes sense to me.  I see how it works.  Thank you for the explanation.

danenbm (approved):
LGTM

PR #80, opened 2023-06-15
fix(ingester): fix compressed nft burn indexing (#24)

NicolasPennie:
# Overview
Fix burn indexing. There is no event for the old bubblegum versions so we index the burn using only the changelog.

# Testing
It's been running in Helius prod for a while now.


danenbm (approved):
LGTM.  Thank  you for fixing this.  I think Redeem may also need a similar update but I can handle that.

PR #84, opened 2023-06-22
fix(api): fix asset sorting

NicolasPennie:
## Overview

Use IndexMap to preserve sorting order when constructing the full asset. This fixes sorting bugs when customers provide non-default sort fields.

## Testing

Running in Helius prod


digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -250,14 +248,11 @@ pub async fn get_assets_by_condition(
     pagination: &Pagination,
     limit: u64,
 ) -> Result<Vec<FullAsset>, DbErr> {
-    let mut stmt = asset::Entity::find().distinct_on([(asset::Entity, asset::Column::Id)]);
+    let mut stmt = asset::Entity::find();
```
danenbm, line 251: Was this causing issues as well, or it it just being removed because it is not needed because `Id` is a primary key so we already know there can only ever be one row per `Id`?

danenbm (approved):
LGTM

PR #87, opened 2023-07-13
Change creator and collection verification to use upserts to support out of order

danenbm:
### Notes
* This PR adds support for creator verification to be indexed with out of order transactions.
* It adds support for collection verification to be indexed with out of order transactions.
* It adds an extra sequence number and a `verified` flag to `asset_grouping` to support out-of-order.

### Testing
* Verified by running a multiple sequences in order and in reverse order.  See table below for details.
* Verified that when a collection is unverified, `getAsset` returns empty grouping.  Then after the collection is verified, `getAsset` returns the collection as a grouping.  This is to remain consistent with existing API behavior.
* See https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/77#issue-1724765232 for more info on overall testing process.

| Transaction sequence | Database result | `getAsset` result | 
| ----------- | ----------- | ----------- |
| `CreateTree`/`Mint`, `VerifyCreator` | OK - `asset`, `asset_creators`, `cl_items` match when instructions run forwards and backwards using PR code.  PR code's `cl_items` table consistent with mainline code when instructions run forwards.  `asset` and `asset_creators` incorrect on main when instructions run backwards. | Did not test this |
| `CreateTree`/`Mint`, `VerifyCollection` | OK - `asset`, `asset_grouping`, `cl_items` match when instructions run forwards and backwards using PR code.  PR code's `asset` and `cl_items` tables consistent with mainline code when instructions run forwards.  `asset_grouping` incorrect on main. | OK - Using PR code, no grouping returned before verified, grouping returned after verified. |
| `CreateTree`/`Mint`, `VerifyCollection`, `UnverifyCollection` | OK - `asset`, `asset_grouping`, `cl_items` match when instructions run forwards and backwards using PR code. | OK - Using PR code, grouping returned after verified, no longer returned after unverified. |
| `CreateTree`/`Mint`, `SetAndVerifyCollection` | `SetAndVerifyCollection` does not parse properly due to Blockbuster error.  Not a regression. | `SetAndVerifyCollection` does not parse properly due to Blockbuster error.  Not a regression. |
| `CreateTree`/`MintToCollection`, `UnverifyCollection` | OK - `asset`, `asset_grouping`, `cl_items `match when instructions run forwards and backwards using PR code. | OK - Using PR code, grouping returned after verified, no longer returned after unverified. |


nft_ingester/src/program_transformers/token_metadata/v1_asset.rs
```diff
@@ -367,8 +367,8 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
                     creator: Set(c.address.to_bytes().to_vec()),
                     share: Set(c.share as i32),
                     verified: Set(c.verified),
-                    seq: Set(0), // do we need this here @micheal-danenberg?
-                    slot_updated: Set(slot_i),
+                    seq: Set(Some(0)), // do we need this here @micheal-danenberg?
```
NicolasPennie, line 370: I feel like you should answer this now 😂

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -233,6 +233,7 @@ where
                 let creators = &metadata.creators;
                 if !creators.is_empty() {
                     let mut db_creators = Vec::with_capacity(creators.len());
+                    let mut db_verified_creators = Vec::with_capacity(creators.len());
```
NicolasPennie, line 236: What's the purpose of this?

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -241,34 +242,59 @@ where
...
+                                asset_creators::Column::Position,
                                 asset_creators::Column::Share,
+                                asset_creators::Column::SlotUpdated,
+                            ])
+                            .to_owned(),
+                        )
+                        .build(DbBackend::Postgres);
+                    txn.execute(query).await?;
+
+                    let mut query = asset_creators::Entity::insert_many(db_verified_creators)
```
NicolasPennie, line 278: So we're re-inserting the same creators again? This code would benefit from some comments to explain this approach. It doesn't seem intuitive (to me at least).

danenbm: The second time we are just actually updated the `verified` and `seq` fields.  The reason the `ActiveModel` used in the second Vec still contains the `asset_id` and the `creator` is really just a matter of locating the record to update. I did it like this to keep it consistent with the rest of the code using the new ordering strategy, which is all using `upserts`. It is also because of the code structure that was there which was building up a Vec and then trying to do a single database call of `insert_many`.

If we think this pattern of `upserting` even when we know the record exists is too inefficient, I could change it around to do just do an `update` to set the `verified` status after the base information is `upserted`. But then I also have to refactor the existing looping structure a bit.

nft_ingester/src/program_transformers/token_metadata/v1_asset.rs
danenbm: We no longer need this as its no longer `NOT NULL` for `asset` and `asset_creators` tables, and it is not used for token-metadata based assets.  So I removed setting it for those tables.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -232,43 +232,79 @@ where
...
                             .update_columns([
-                                asset_creators::Column::Creator,
+                                asset_creators::Column::Position,
                                 asset_creators::Column::Share,
+                                asset_creators::Column::SlotUpdated,
+                            ])
+                            .to_owned(),
+                        )
+                        .build(DbBackend::Postgres);
+                    txn.execute(query).await?;
```
linuskendall, line 283: does this query need a where check for example on SlotUpdated?

linuskendall (approved):
one comment on the updating of position

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
linuskendall: if it cannot be reordered, then is the update needed atg all?

danenbm: It is still needed to be an update because if 'verify_creator` is processed first, then there will be a record but it won't actually have all this base information from the time of minting.

danenbm: > does this query need a where check for example on SlotUpdated?

It does not need protection because the base creator info is ONLY filled in at time of mint.

danenbm: We will likely be updating this to support changes to creator array and collection.

PR #88, opened 2023-07-17
Order by creator position

linuskendall:
Order by creator position so that hashing of the asset metadata generates predictable results.


danenbm (approved):

danenbm:
CI was fixed in main

PR #90, opened 2023-07-20
Fix collection verification and change to use upserts to support out of order processing

danenbm:
### Notes
* This PR is based on https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/77 and https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/87
* It adds support for collection verification to be indexed with out of order transactions.
* It also adds a `verified` flag so that collection verification is indexed properly.

### Testing
* Verified by running a multiple sequences in order and in reverse order, and either way get same end result in `asset` and `asset_grouping` tables.
* Verified that when a collection is unverified, `getAsset` returns empty grouping.  Then after the collection is verified, `getAsset` returns the collection as a grouping.  This is to remain consistent with existing API behavior.
* See https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/77#issue-1724765232 for more info on overall testing process.

| Transaction sequence | Database result | `getAsset` result | 
| ----------- | ----------- | ----------- |
| `CreateTree`/`Mint`, `VerifyCollection` | OK - `asset`, `asset_grouping`, `cl_items` match when PR code run forwards and backwards.  PR code's `asset` and `cl_items` tables consistent with mainline code run forwards.  `asset_grouping` incorrect on main. | OK - Using PR code, no grouping returned before verified, grouping returned after verified. |
| `CreateTree`/`Mint`, `VerifyCollection`, `UnverifyCollection` | OK - `asset`, `asset_grouping`, `cl_items` match when PR code is run forwards and backwards. | OK - Using PR code, grouping returned after verified, no longer returned after unverified. |
| `CreateTree`/`Mint`, `SetAndVerifyCollection` | `SetAndVerifyCollection` does not parse properly due to Blockbuster error.  Not a regression. | `SetAndVerifyCollection` does not parse properly due to Blockbuster error.  Not a regression. |
| `CreateTree`/`MintToCollection`, `UnverifyCollection` | OK - `asset`, `asset_grouping`, `cl_items `match when PR code is run forwards and backwards. | OK - Using PR code, grouping returned after verified, no longer returned after unverified. |

### DIFF
--- a/digital_asset_types/src/dao/scopes/asset.rs
+++ b/digital_asset_types/src/dao/scopes/asset.rs
@@ -65,7 +65,8 @@ pub async fn get_grouping(
         .filter(
             Condition::all()
                 .add(asset_grouping::Column::GroupKey.eq(group_key))
-                .add(asset_grouping::Column::GroupValue.eq(group_value)),
+                .add(asset_grouping::Column::GroupValue.eq(group_value))
+                .add(asset_grouping::Column::Verified.eq(true)),
         )
         .count(conn)
         .await?;
@@ -193,7 +194,7 @@ pub async fn get_related_for_assets(
         {
             let id = asset.id.clone();
             let fa = FullAsset {
-                asset: asset,
+                asset,
                 data: ad.clone(),
                 authorities: vec![],
                 creators: vec![],
@@ -229,6 +230,7 @@ pub async fn get_related_for_assets(
 
     let grouping = asset_grouping::Entity::find()
         .filter(asset_grouping::Column::AssetId.is_in(ids.clone()))
+        .filter(asset_grouping::Column::Verified.eq(true))
         .order_by_asc(asset_grouping::Column::AssetId)
         .all(conn)
         .await?;
@@ -289,6 +291,7 @@ pub async fn get_by_id(
         .await?;
     let grouping: Vec<asset_grouping::Model> = asset_grouping::Entity::find()
         .filter(asset_grouping::Column::AssetId.eq(asset.id.clone()))
+        .filter(asset_grouping::Column::Verified.eq(true))
         .all(conn)
         .await?;
     Ok(FullAsset {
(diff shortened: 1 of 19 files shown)


digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -83,7 +83,8 @@ pub async fn get_by_grouping(
 ) -> Result<Vec<FullAsset>, DbErr> {
     let condition = asset_grouping::Column::GroupKey
         .eq(group_key)
-        .and(asset_grouping::Column::GroupValue.eq(group_value));
+        .and(asset_grouping::Column::GroupValue.eq(group_value))
+        .and(asset_grouping::Column::Verified.eq(true));
```
linuskendall, line 87: is this condition strictly required? seems it changes the query a bit ?

linuskendall (approved):

digital_asset_types/src/dao/scopes/asset.rs
danenbm: So the way it used to be was that stuff was ONLY in the `asset_grouping` table if it was `verified`.  Because of that, minting with an unverified collection and verifying later would not index properly (had nothing to do with order of ingestion. The ingester just ignored the instructions).

So adding the verified flag is needed to make the indexing work, and changing the query makes it consistent with how it worked before.  It will not return an unverified collection, just as before.

danenbm: I will add this check for others as well.  We can change the API to instead return the verified value as creators are returned, or we can add a bool flag for the caller to request verified only.  But if so, I think that should be a follow on PR to discuss specifically.

The reason I don't want to automatically include unverified collections is that it could lead to issues with marketplaces listing something that is not truly part of a collection since anyone could set their collection to anything without it being verified.

PR #91, opened 2023-07-25
Adds a reindex param

linuskendall:
The reindex param is used to indicate the need for reindexing. This fixes an issue where the reindexing procedure would reset the field to processing, removing the stale but valid data that already existed in the column. It also makes it easier to find all the assets that require reindexing by having an indexed field to search for.


danenbm:
This should be covered by: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/110

linuskendall:
This was lumped in with other changes

PR #93, opened 2023-07-25
Remove trims

linuskendall:
The data returned should be identical to that on chain. Trims modify this in a way that breaks the compatibility. Remove trim.


digital_asset_types/src/json/chain_data.rs
```diff
@@ -20,7 +20,7 @@ pub struct ChainDataV1 {
 
 impl ChainDataV1 {
     pub fn sanitize(&mut self) {
-        self.name = self.name.trim().replace("\0", "").to_string();
-        self.symbol = self.symbol.trim().replace("\0", "").to_string();
+        self.name = self.name.replace('\0', "");
```
nikhil-helius, line 23: Can the onchain contain `\0`, for example, "John\0Doe"? Wouldn't removing `\0` cause a different with onchain?

linuskendall: We reviewed this and as far as we can see these are null terminated strings everywhere so John\0Doe would be invalid anyway. That was the conclusion we came to when we checked this - and also should someone put string like this on chain it would cause a lot of trouble elsewhere.

nikhil-helius: I see. We see an asset with the symbol `CAMP\0\0\0\0\0\0`.  This could potentially cause a different metadata hash where it is computed on the raw data as opposed to the one which is rendered by the api. If someone computes the hash based on what the api returns and matches with the data hash, its going to be different

linuskendall:
That's an interesting asset indeed... I think that technically breaks a
bunch of the apis and toolings around the ecosystem, but it may indeed be
wise to be able to represent this in raw format. I see you added this to
the API already but not upstreamed yet. However, I wonder if there's also a
question if this kind of asset is attempting to be malicious or create
problems?

digital_asset_types/src/json/chain_data.rs
linuskendall: If we have a sample of such an asset maybe we should try to compute the metadata differently. If this is every transferred into a string even during the computation of metadata then the additional null characters should be ignored.

nikhil-helius: This is the asset `9cA3ocsaxb52HWjRaQsp1M7xqD5wqyvK9wJ6xHygY3ku` and this is the corresponding mint tx `4ht8uMYFfp97NxqyKJ9pjbnUubN1WopruoXQrUt1392Y8bX1EsUDH7sywkyu7MS5vLv7q6uCDE4VL2uXTFk5Xpzm` which you can replay to see the above symbol.  We would have to change the method for the way we hash and recompute the hash for all the assets to fix the discrepancy

danenbm: Another proposal by @lorisleiva was to include the metadata with no processing/clean up under the compression section, to make it easier for users to get the exact metadata needed for instructions such as `verify_creator` that take `MetadataArgs` as an argument.

Then maybe the normal API could return a trimmed string, and under compression.metadata it could be the raw untrimmed version which should help with hashing?  Just a thought, I know it doesn't resolve the talk about \0

danenbm:
> That's an interesting asset indeed... I think that technically breaks a bunch of the apis and toolings around the ecosystem, but it may indeed be wise to be able to represent this in raw format. I see you added this to the API already but not upstreamed yet. However, I wonder if there's also a question if this kind of asset is attempting to be malicious or create problems?

For the purposes of this PR change should we just remove BOTH the `trim()` AND the `\0` replace?

digital_asset_types/src/json/chain_data.rs
linuskendall: I have added #117  to add raw_name and raw_symbol. This change removing the trim is 100% needed though, so if we want to do further refactoring we can probably remove the `name` and `symbol` and just use `raw_name` and `raw_symbol`. But for now I suggest that we merge this PR and make a new PR for the bigger change.

linuskendall:
We need to keep the \0 filter because we are serialising this to JSONB in postgres and postgres doesn't seem to allow null characters in jsonb:
```
psql (13.12 (Ubuntu 13.12-1.pgdg20.04+1))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

plerkle=> SELECT '{"a_json_object": "with_a_\u0000_byte"}'::jsonb;
ERROR:  unsupported Unicode escape sequence
LINE 1: SELECT '{"a_json_object": "with_a_\u0000_byte"}'::jsonb;
```

More details: https://www.commandprompt.com/blog/null-characters-workarounds-arent-good-enough/

linuskendall:
Rust and JS can handle `\0` in the name, but Postgres cannot.

danenbm (approved):

PR #94, opened 2023-07-25
Add configurable cl audits table

linuskendall:
This adds the cl_audits table which contains all the transactions for a specific compressed asset. This has some quite heavy increased data usage, so I have added a configuration param for this.

I think helius-labs have used this same table for a new rpc call they added so getting this merged in would help upstream those changes too later.


NicolasPennie:
Good call. I agree that this should be configurable since it's quite intensive. We'll merge our endpoint, too. It's experimental at the moment but if it's controlled via config I think it makes sense to include as an optional endpoint.

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -75,6 +90,12 @@ where
         txn.execute(query)
             .await
             .map_err(|db_err| IngesterError::StorageWriteError(db_err.to_string()))?;
+
+
+        // Insert the audit item after the insert into cl_items have been completed
```
danenbm, line 95: Wouldn't you want insert to the audit table _before_ inserting to `cl_items`, so that if the insert to `cl_items` fails, it remains in the audit table and is audited and fixed later?

danenbm (approved):
LGTM but left a question.

NicolasPennie (approved):

PR #99, opened 2023-07-26
Updated tools/tests

linuskendall:
Adding updates to the tool/tests done on Triton's branch. This adds new tools for bgtask creation and for tree status.


tests/tree-status/src/main.rs
```diff
@@ -0,0 +1,820 @@
...
+
+#[derive(Subcommand, Clone)]
+enum Action {
+    /// Checks a single merkle tree to check if it's fully indexed
+    CheckTree {
+        #[arg(short, long)]
+        pg_url: String,
+        #[arg(short, long, help = "Tree pubkey")]
+        tree: String,
+    },
```
danenbm, line 195: Maybe its fine because all these action args tell you what they are for, but it was really nice in the bgtask main to have a header kind of saying overall what the tool was for and how it was used in practice.

So a minor request would be to add a similar header to tree-status and potentially the others.  Although this is just a "nice to have" and not a blocker by any means, its already much cleaned up overall.

tests/txn_forwarder/src/main.rs
```diff
@@ -1,144 +1,348 @@
...
+    async fn send(
+        &self,
+        pubkey: Option<Pubkey>,
+        signature: Signature,
+        tx: EncodedConfirmedTransactionWithStatusMeta,
+    ) -> anyhow::Result<()> {
+        // ignore if tx failed or meta is missed
+        let meta = tx.transaction.meta.as_ref();
+        if meta.map(|meta| meta.status.is_err()).unwrap_or(true) {
+            return Ok(());
```
danenbm, line 135: Do you want to print a warning if there's an error but still return Ok?

danenbm (approved):
This all looks awesome to me, thank you for cleaning things up and adding features to existing tools and the new tools!

My assumption is that these tools are all used today and thus tested by production usage.

PR #101, opened 2023-07-26
Add cargo workspace

linuskendall:
Adding a cargo workspace file.


NicolasPennie (approved):

danenbm (approved):
LGTM.  Looks like it has some of the changes already from 99 so I'm sure it will be fixed by resolving merge conflicts, but this is good to add the workspace organization!

PR #102, opened 2023-07-27
Add new backfill queues

linuskendall:
This moves the queues the acc forwarder and txn forwarder uses to be backfill queues. This allows us to backfill without disturbing ingestion of new data coming in. This can be quite important considering the need for latest data for proofs to work. It aligns well with the changes introduced via out of order processing allowing us to introduce past history while still maintaining proof status.


tools/txn_forwarder/Cargo.toml
```diff
@@ -13,8 +13,8 @@ flatbuffers = "23.1.21"
 futures = "0.3.28"
 lazy_static = "1.4.0"
 log = "0.4.17"
-plerkle_messenger = { version = "1.5.0", features = ["redis"] }
-plerkle_serialization = "1.5.0"
+plerkle_serialization = { git = "https://github.com/metaplex-foundation/digital-asset-validator-plugin.git", rev = "35a8801c" }
```
danenbm, line 16: You could probably just use `plerkle_serialization = { version = "1.5.2" }` since that's what we had to leave it in the ingester due to blockbuster using that version... But its fine to use the pinned version as well.

tools/acc_forwarder/Cargo.toml
```diff
@@ -15,8 +15,8 @@ futures = "0.3.28"
 lazy_static = "1.4.0"
 log = "0.4.17"
 mpl-token-metadata = "1.8.3"
-plerkle_messenger = { version = "1.5.2", features = ["redis"] }
-plerkle_serialization = "1.5.2"
+plerkle_serialization = { git = "https://github.com/metaplex-foundation/digital-asset-validator-plugin.git", rev = "35a8801c" }
```
danenbm, line 18: You could probably just use `plerkle_serialization = { version = "1.5.2" }` since that's what we had to leave it in the ingester due to blockbuster using that version... But its fine to use the pinned version as well.

danenbm (approved):
LGTM as discussed offline request small 1-2 sentence description of the perf improvement.

PR #122, opened 2023-09-22
Fix raw name build

linuskendall:
Fixes the build of migrator.


Juanito87 (approved):
LGTM, this fixes the build issue.

PR #124, opened 2023-09-22
Updating test workflow

Juanito87:
Updates taest workflow to build das-api components, and validate that a pr works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,24 +41,23 @@ jobs:
               ${{ matrix.os }}_digital-asset-rpc-infrastructure
 
       - name: build digital asset rpc infra
-        working-directory: digital-asset-rpc-infrastructure/
         run: cargo build --verbose --release
 
       - name: build das_api
-        working-directory: digital-asset-rpc-infrastructure/das_api
+        working-directory: das_api
         run: cargo build --verbose --release
 
       - name: build migration
-        working-directory: digital-asset-rpc-infrastructure/migration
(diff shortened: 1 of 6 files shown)


Juanito87:
Current build process will fail untill pr #122 is merged (that 's the fix for the migration binary compile error).

.github/workflows/test.yml
```diff
@@ -1,31 +1,118 @@
...
+name: Build das api components
+# This workflow uses github runners.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+# This may be adjusted to whatever suits best your runners config.
+# Current config will build on manual trigger or pull-request (each push)
 on:
-  push:
```
danenbm, line 3: So just to confirm, what this will do is not run the CI when you are pushing to your dev branch, but then as soon as you open a PR to any branch, it will run on the PR branch every time there's a check in.  That seems like the right behavior.

But should we add something like:
```
push:
    branches: [main]
```
so that CI will run in the case that we do a hotfix directly to main without a PR?

danenbm (changes requested):

.github/workflows/test.yml
Juanito87: Ok, if later on more branches are needed,it's simple to expand the list.
If there is any other testing scenario yoou may need, let me know.

danenbm (approved):
LGTM thank you for doing this, it is a good improvement!!!

PR #126, opened 2023-09-28
Switch to github runners

febo:
Switch runners from buildjet to github.


febo:
Closed in favour of #124

PR #127, opened 2023-10-09
Update to Solana 1.16 and use new Bubblegum Rust client

danenbm:
### Overview
Updates DAS API to use Solana 1.16 versions, plus the initial changes for Bubblegum Update Metadata (uses the Bubblegum Rust client that supports Update Metadata, but doesn't actually do anything with it yet).

### Details
* Change path for load generation
* Update to Solana 1.16 versions 
  * Update to Solana 1.16.16 (version that does not have tokio pinned and solana-transaction-status supports Borsh 0.10.3).
  * Update blockbuster to latest branch that supports Bubblegum Update Metadata instruction parsing (still a commit ID).
  * Update to Borsh 0.10.3.
  * Update to Plerkle Messenger 1.6.0.
  * Update to Bubblegum 1.0.1-beta.2 (Rust client version that supports Update Metadata and limits to Borsh < 1.0.0).
  * Update to SPL Account Compression 0.2.0.
  * Update to Anchor 0.28.0.
  * Update Candy Machine and Candy Guard to 2.0.x.
  * spl-associated-token-account and spl-token use same version ranges as Bubblegum program crate.
  * Update SPL No-op.
  * MPL Token Metadata exact version: 2.0.0-beta.1.

### DIFF
--- a/Load.Dockerfile
+++ b/Load.Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.70-bullseye AS chef
+FROM rust:1.73-bullseye AS chef
 RUN cargo install cargo-chef
 FROM chef AS planner
 COPY tools/load_generation /rust/load_generation/
@@ -17,7 +17,7 @@ RUN cargo chef cook --release --recipe-path recipe.json
 COPY tools/load_generation .
 # Build application
 RUN cargo build --release
-FROM rust:1.70-slim-bullseye
+FROM rust:1.73-slim-bullseye
 ARG APP=/usr/src/app
 RUN apt update \
     && apt install -y curl ca-certificates tzdata \
--- a/Load.Dockerfile
+++ b/Load.Dockerfile
@@ -17,7 +17,7 @@ RUN cargo chef cook --release --recipe-path recipe.json
 COPY tools/load_generation .
 # Build application
 RUN cargo build --release
-FROM rust:1.63-slim-bullseye
+FROM rust:1.70-slim-bullseye
 ARG APP=/usr/src/app
 RUN apt update \
     && apt install -y curl ca-certificates tzdata \
--- a/Load.Dockerfile
+++ b/Load.Dockerfile
@@ -1,20 +1,20 @@
 FROM rust:1.70-bullseye AS chef
 RUN cargo install cargo-chef
 FROM chef AS planner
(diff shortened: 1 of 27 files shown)


Load.Dockerfile
```diff
@@ -1,23 +1,23 @@
...
 # Build dependencies - this is the caching Docker layer!
-COPY tests/load_generation/Cargo.toml .
+COPY tools/load_generation/Cargo.toml .
 RUN cargo chef cook --release --recipe-path recipe.json
-COPY tests/load_generation .
+COPY tools/load_generation .
 # Build application
 RUN cargo build --release
-FROM rust:1.63-slim-bullseye
+FROM rust:1.70-slim-bullseye
```
fanatid, line 20: is there any reason do not updating to the latest stable (1.73)?

danenbm: will update

danenbm: done

nft_ingester/src/config.rs
```diff
@@ -71,17 +71,15 @@ impl IngesterConfig {
     }
 
     pub fn get_account_backfill_stream_worker_count(&self) -> u32 {
-        self.account_backfill_stream_worker_count
-            .unwrap_or_else(|| self.get_account_stream_worker_count())
+        self.account_backfill_stream_worker_count.unwrap_or(0)
```
danenbm, line 74: Have to set this to zero for now because we are not configuring any backfill streams in the plugin.

nft_ingester/src/config.rs
```diff
@@ -71,17 +71,15 @@ impl IngesterConfig {
...
     }
 
     pub fn get_transaction_stream_worker_count(&self) -> u32 {
         self.transaction_stream_worker_count.unwrap_or(2)
     }
 
     pub fn get_transaction_backfill_stream_worker_count(&self) -> u32 {
-        self.transaction_backfill_stream_worker_count
-            .unwrap_or_else(|| self.get_transaction_stream_worker_count())
+        self.transaction_backfill_stream_worker_count.unwrap_or(0)
```
danenbm, line 82: Have to set this to zero for now because we are not configuring any backfill streams in the plugin.

NicolasPennie (approved):

PR #130, opened 2023-10-15
fix: allow no metadata for cnft mints

NicolasPennie:
# Overview

Bubblegum allows for empty URIs. When this happens, the indexer will reject the transaction leading to a missing NFT in the API response. To fix, we allow the record to index and skip the off-chain indexing update.

# Testing
Running in Helius fork in prod.
Keep in mind that we are 1.14 for ingester at the moment, so I have not perfectly reproduced these changes on the Metaplex version. The changes are simple though and should be fine.


danenbm (approved):
LGTM

PR #133, opened 2023-10-17
Upstream Helius features

tahsintunan:
### Features:
- Add a new param `displayOptions` which can be used to opt-in for additional info.
- Allow users to see unverified collections via `displayOptions` opt-in.  `getAssetsByGroup` and `searchAssets` will return unverified collections as well if this flag is set.
- Add two new endpoints: `getAssetBatch` and `getAssetProofBatch`.
- New cursor-based keyset pagination method. The cursor is the same as the asset ID and behaves like the  `after` option. When specified without any page options, it defaults to the cursor.
- Add `token_standard` in DAS response.
- Add support for searching by asset `name` in the `searchAssets` function. The `owner_address` field must be provided in conjunction with the `name` field.

### Testing
- Currently running on Helius prod.

### DIFF
--- a/das_api/src/builder.rs
+++ b/das_api/src/builder.rs
@@ -22,17 +22,16 @@ impl RpcApiBuilder {
         })?;
         module.register_alias("getAssetProof", "get_asset_proof")?;
 
-        module.register_async_method(
-            "get_asset_proof_batch",
-            |rpc_params, rpc_context| async move {
-                let payload = rpc_params.parse::<GetAssetProofBatch>()?;
-                rpc_context
-                    .get_asset_proof_batch(payload)
-                    .await
-                    .map_err(Into::into)
-            },
-        )?;
-        module.register_alias("getAssetProofBatch", "get_asset_proof_batch")?;
+        module.register_async_method("get_asset_proofs", |rpc_params, rpc_context| async move {
+            let payload = rpc_params.parse::<GetAssetProofs>()?;
+            rpc_context
+                .get_asset_proofs(payload)
+                .await
+                .map_err(Into::into)
+        })?;
+        module.register_alias("getAssetProofs", "get_asset_proofs")?;
+        module.register_alias("get_asset_proof_batch", "get_asset_proofs")?;
+        module.register_alias("getAssetProofBatch", "get_asset_proofs")?;
 
         module.register_async_method("get_asset", |rpc_params, rpc_context| async move {
             let payload = rpc_params.parse::<GetAsset>()?;
@@ -40,14 +39,13 @@ impl RpcApiBuilder {
         })?;
         module.register_alias("getAsset", "get_asset")?;
 
-        module.register_async_method("get_asset_batch", |rpc_params, rpc_context| async move {
-            let payload = rpc_params.parse::<GetAssetBatch>()?;
-            rpc_context
-                .get_asset_batch(payload)
-                .await
-                .map_err(Into::into)
+        module.register_async_method("get_assets", |rpc_params, rpc_context| async move {
+            let payload = rpc_params.parse::<GetAssets>()?;
+            rpc_context.get_assets(payload).await.map_err(Into::into)
         })?;
-        module.register_alias("getAssetBatch", "get_asset_batch")?;
+        module.register_alias("getAssets", "get_assets")?;
+        module.register_alias("get_asset_batch", "get_assets")?;
+        module.register_alias("getAssetBatch", "get_assets")?;
 
         module.register_async_method(
             "get_assets_by_owner",
--- a/das_api/src/builder.rs
+++ b/das_api/src/builder.rs
@@ -22,11 +22,32 @@ impl RpcApiBuilder {
         })?;
         module.register_alias("getAssetProof", "get_asset_proof")?;
 
+        module.register_async_method(
+            "get_asset_proof_batch",
+            |rpc_params, rpc_context| async move {
+                let payload = rpc_params.parse::<GetAssetProofBatch>()?;
+                rpc_context
+                    .get_asset_proof_batch(payload)
+                    .await
+                    .map_err(Into::into)
+            },
+        )?;
+        module.register_alias("getAssetProofBatch", "get_asset_proof_batch")?;
+
         module.register_async_method("get_asset", |rpc_params, rpc_context| async move {
             let payload = rpc_params.parse::<GetAsset>()?;
             rpc_context.get_asset(payload).await.map_err(Into::into)
         })?;
         module.register_alias("getAsset", "get_asset")?;
+
+        module.register_async_method("get_asset_batch", |rpc_params, rpc_context| async move {
+            let payload = rpc_params.parse::<GetAssetBatch>()?;
+            rpc_context
+                .get_asset_batch(payload)
+                .await
+                .map_err(Into::into)
+        })?;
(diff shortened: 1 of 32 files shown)


NicolasPennie (approved):

nikhil-helius (approved):

danenbm (changes requested):
Overall nice clean changes overall had one very minor suggestion and I think @febo had a comment about the new method names.

digital_asset_types/src/rpc/response.rs
```diff
@@ -30,6 +30,8 @@ pub struct AssetList {
     pub before: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub after: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cursor: Option<String>,
```
febo, line 34: The PR description mentions "`cursor` behaves like the  `after` option", so I was wondering why it is necessary to add `cursor`. From a devx perspective, it would be better to a single way to achieve something in the API, so if `cursor` and `after` are the same, then we should keep one of them.

digital_asset_types/src/rpc/display_options.rs
```diff
@@ -0,0 +1,9 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq, JsonSchema, Default)]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct DisplayOptions {
+    #[serde(default)]
+    pub show_unverified_collections: bool,
```
febo, line 8: Including unverified collections is not strictly a "display option" since it changes the underline query. Perhaps we should add `groupVerifiedOnly` top-level parameter to the handlers, which by default is `true` but can be set to `false`. Also, note that "collection" is one type of "group", so we should try to keep the names more generic.

digital_asset_types/src/dapi/get_asset.rs
```diff
@@ -1,10 +1,33 @@
...
+    db: &DatabaseConnection,
+    id: Vec<u8>,
+    display_options: &DisplayOptions,
+) -> Result<Asset, DbErr> {
     let asset = scopes::asset::get_by_id(db, id, false).await?;
-    asset_to_rpc(asset)
+    asset_to_rpc(asset, display_options)
+}
+
+pub async fn get_asset_batch(
```
febo, line 18: We need to keep a name pattern for handlers, so the API does not look over-complicated. So far we have `getAssets*` methods that return multiple assets – ideally we should keep with this, so `get_asset_batch` would be names `get_assets`.

digital_asset_types/src/dapi/change_logs.rs
```diff
@@ -61,48 +74,169 @@ pub async fn get_proof_for_asset(
...
+        leaf.tree,
+        leaf.node_idx,
+        leaf.hash,
+        &req_indexes,
+        &required_nodes,
+    );
+    Ok(asset_proof)
+}
+
+pub async fn get_asset_proof_batch(
```
febo, line 92: Same comment as the one on `get_asset_batch`. We have been using an extra `s` to indicate methods that return multiple items. Therefore, we should consider naming this `get_asset_proofs`, assuming the `AssetProof` is an entity, so the method will return a list of `AssetProof`s.

das_api/src/builder.rs
```diff
@@ -22,12 +22,33 @@ impl RpcApiBuilder {
...
+            "get_asset_proof_batch",
+            |rpc_params, rpc_context| async move {
+                let payload = rpc_params.parse::<GetAssetProofBatch>()?;
+                rpc_context
+                    .get_asset_proof_batch(payload)
+                    .await
+                    .map_err(Into::into)
+            },
+        )?;
+        module.register_alias("getAssetProofBatch", "get_asset_proof_batch")?;
```
febo, line 35: See comment about the name on the method implementation.

febo (changes requested):
Thanks for this. The implementation looks very nice. Left a few comments, mostly related to the naming. We should try to have a consistent API to improve devx. Currently we are using the pattern of adding an `s` at the end to represent cases where multiple assets are returned – e.g., `get_assets_by_authority`. It would be good if we can follow a similar pattern.

das_api/src/builder.rs
```diff
@@ -22,12 +22,33 @@ impl RpcApiBuilder {
...
         module.register_alias("getAsset", "get_asset")?;
 
+        module.register_async_method("get_asset_batch", |rpc_params, rpc_context| async move {
+            let payload = rpc_params.parse::<GetAssetBatch>()?;
+            rpc_context
+                .get_asset_batch(payload)
+                .await
+                .map_err(Into::into)
+        })?;
+        module.register_alias("getAssetBatch", "get_asset_batch")?;
```
febo, line 50: See comment about the name on the method implementation.

digital_asset_types/src/rpc/response.rs
NicolasPennie: Discussed on a call. The main reason is for simplicity / UX: a developer does not want to have to re-think what before/after means as they adjust the sort order. Its far easier to take a token/cursor and fire-n-forget. 

However, you can use before/after in combination with before/after if you wanted to. 
E.g. scope down a query to a range and then paginate throughout that range.

digital_asset_types/src/rpc/display_options.rs
NicolasPennie: We didn't know if all groups would have a form of verification or not. You might also want to see some groups verified, some not.

The key detail here is that we leverage "show_unverified_collections" across multiple endpoints. So `groupVerifiedOnly` is confusing depending on the API.

I'll propose an option:
What if we just rename "displayOptions" to "options" like we discussed on the call?

digital_asset_types/src/dapi/get_asset.rs
NicolasPennie: We can add an alias if people feel strongly about this. I don't want to break anything for existing Helius users. We have quite a bit of traffic on these endpoints.

digital_asset_types/src/rpc/display_options.rs
febo: The `options` sounds good.

digital_asset_types/src/dapi/get_asset.rs
febo: I think we agreed to maintain the consistency with existing methods on the call. So we would have `get_assets` and `get_asset_proofs`. The alias seems a good solutions to not break existing Helius users.

digital_asset_types/src/rpc/display_options.rs
NicolasPennie: Correct.

digital_asset_types/src/dapi/get_asset.rs
tahsintunan: Addressed

digital_asset_types/src/rpc/display_options.rs
tahsintunan: Addressed

danenbm (approved):
This LGTM.  I asked @febo to take a look and approve as well because he had the most recent comments.

febo (approved):
LGTM.

NicolasPennie (approved):

PR #135, opened 2023-10-23
Danenbm/update metadata parsing 2

danenbm:
WIP second version


blockiosaurus (approved):
This looks good based on my very little understanding of DAS

nft_ingester/src/tasks/common/mod.rs
```diff
@@ -110,24 +111,29 @@ impl BgTask for DownloadMetadataTask {
...
-                    self.name(),
-                    db
-                ))
-            })?;
+        let mut query = asset_data::Entity::update(model)
+            .filter(asset_data::Column::Id.eq(download_metadata.asset_data_id.clone()));
+        if download_metadata.seq != 0 {
+            query = query.filter(
+                Condition::any()
+                    .add(asset_data::Column::DownloadMetadataSeq.lt(download_metadata.seq))
```
danenbm, line 126: Note I think this should be `lteq`

danenbm: fixed

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
```diff
@@ -0,0 +1,292 @@
...
+                    Some(true),
+                    name.into_bytes().to_vec(),
+                    symbol.into_bytes().to_vec(),
+                )
+                .await?;
+
+                // Upsert into `asset` table.
+
+                // Set base mint info.
+                let tree_id = bundle.keys.get(5).unwrap().0.to_vec();
```
NicolasPennie, line 151: This value should be parsed from blockbuster instead of via an array index

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
```diff
@@ -0,0 +1,292 @@
...
+                    delegate,
+                    seq as i64,
+                )
+                .await?;
+
+                upsert_asset_with_seq(txn, id_bytes.to_vec(), seq as i64).await?;
+
+                // Upsert into `asset_v1_account_attachments` table.
+                let (edition_attachment_address, _) = find_master_edition_account(&id);
+                unprotected_upsert_asset_v1_account_attachments(
```
NicolasPennie, line 217: Does account attachments do anything right now?

Assuming it does, why is it safe to do "unprotected" here?

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -86,92 +89,38 @@ where
...
-                    specification_asset_class: Set(Some(SpecificationAssetClass::Nft)),
-                    nonce: Set(Some(nonce as i64)),
-                    royalty_target_type: Set(RoyaltyTargetType::Creators),
-                    royalty_target: Set(None),
-                    royalty_amount: Set(metadata.seller_fee_basis_points as i32), //basis points
-                    asset_data: Set(Some(id_bytes.to_vec())),
-                    slot_updated: Set(Some(slot_i)),
-                    ..Default::default()
-                };
+                unprotected_upsert_asset_data(
```
NicolasPennie, line 92: Is unprotected only possible here because of `!asset_should_be_updated` from earlier?

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -86,92 +89,38 @@ where
...
-                // Do not overwrite changes that happened after the asset was decompressed.
-                query.sql = format!(
-                    "{} WHERE excluded.slot_updated > asset.slot_updated OR asset.slot_updated IS NULL",
-                    query.sql
-                );
-                txn.execute(query)
-                    .await
-                    .map_err(|db_err| IngesterError::AssetIndexError(db_err.to_string()))?;
+                // Set base mint info.
+                let tree_id = bundle.keys.get(3).unwrap().0.to_vec();
```
NicolasPennie, line 110: Can this come from blockbuster instead?

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -211,121 +164,36 @@ where
...
-                        )
-                        .build(DbBackend::Postgres);
-                    query.sql = format!(
-                        "{} WHERE excluded.seq > asset_creators.seq OR asset_creators.seq IS NULL",
-                        query.sql
-                    );
-                    txn.execute(query).await?;
-                }
+                // Upsert into `asset_creators` table.
+                unprotected_upsert_creators(
```
NicolasPennie, line 177: I don't think we can do an unprotected upsert here. What happens if there's an out-of-order update with creator verify/unverify?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -441,3 +441,327 @@ where
...
+                    asset_creators::Column::Creator,
+                ])
+                .update_columns([
+                    asset_creators::Column::Verified,
+                    asset_creators::Column::VerifiedSeq,
+                ])
+                .to_owned(),
+            )
+            .build(DbBackend::Postgres);
+        query.sql = format!(
```
NicolasPennie, line 696: Why is the method called "unprotected" if this case exists?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -441,3 +441,327 @@ where
...
+
+pub async fn asset_should_be_updated<T>(
+    txn: &T,
+    id: Vec<u8>,
+    seq: Option<i64>,
+) -> Result<bool, IngesterError>
+where
+    T: ConnectionTrait + TransactionTrait,
+{
+    if let Some(asset) = asset::Entity::find_by_id(id).one(txn).await? {
```
NicolasPennie, line 567: I don't think this approach works. It can lead to race conditions.

Say we go to seq 1 -> 2 with our update

1. Fetch asset, seq says 1
2. Determine it can be updated because 2 > 1
3. Concurrent process upserts new data with seq 3
4. We then revert from seq 3 -> 2

This function can act as a filter, but not a true safety mechanism. It should be clearly labelled as such. Therefore no method this is called should be "unprotected". 

To give an example as a filter:
If current seq = 5 and we want to update to 4, then we know we can eagerly discord the txn.

nft_ingester/src/program_transformers/bubblegum/delegate.rs
```diff
@@ -30,6 +30,13 @@ where
                 ..
             } => {
                 let id_bytes = id.to_bytes();
+
+                // First check to see if this asset has been decompressed or updated by
+                // `update_metadata`.
```
NicolasPennie, line 35: Does update_metadata impact the owner/delegate info? Should this check actually be here?

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -441,3 +441,327 @@ where
...
+    );
+
+    txn.execute(query)
+        .await
+        .map_err(|db_err| IngesterError::StorageWriteError(db_err.to_string()))?;
+
+    Ok(())
+}
+
+pub async fn unprotected_upsert_creators<T>(
```
linuskendall, line 619: this method seems to be doing updates, so shouldn't it be protected?

nft_ingester/src/tasks/common/mod.rs
```diff
@@ -110,24 +111,29 @@ impl BgTask for DownloadMetadataTask {
             id: Unchanged(download_metadata.asset_data_id.clone()),
             metadata: Set(body),
             reindex: Set(Some(false)),
+            download_metadata_seq: Set(Some(download_metadata.seq)),
```
linuskendall, line 114: I am not sure about this addition. I can see that the download_metadata uri should be protected by a seq since that is on chain. But the offchain metadata can change at any time. To me the system should always download the latest metadata that is offchain irrespectively of the seq?

So as long as we know that the URI in the asset_data table is only ever updated to the latest we shouldn't protect the download metadata itself? 

If the URI is updated, then indeed we must redownload metadata but I also don't see a harm in scheduling additional metadata downloads in between. Not having an additional sequence number to check might be quite good also.

Adding this sequence number has limited value and causes extra DB queries/complexity.

linuskendall: If we want to limit the number of downloads, we could either

a) keep track of URI changes and only trigger a redownload if the URI hasn't been downloaded yet
b) make the download metadata task smarter to understand if the offchain data has changed (i.e. do some kind of HEAD check to find the HTTP headers to show if data is changed)

linuskendall (approved):
In principle it looks ok but Nick's concern about the serialisability of the changes rquires addressing. This can be addressed via DB transaction and setting the appropraite isolation levels.

migration/src/m20231019_120101_add_seq_numbers_bgum_update_metadata.rs
```diff
@@ -0,0 +1,79 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .get_connection()
+            .execute(Statement::from_string(
+                DatabaseBackend::Postgres,
+                "
+                ALTER TABLE asset_creators
+                RENAME COLUMN seq to verified_seq;
```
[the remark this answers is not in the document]
danenbm: Is changing the name OK then, given the way you all run migrations?

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
```diff
@@ -0,0 +1,292 @@
...
+                // Set base mint info.
+                let tree_id = bundle.keys.get(5).unwrap().0.to_vec();
+                let seller_fee_basis_points =
+                    if let Some(seller_fee_basis_points) = update_args.seller_fee_basis_points {
+                        seller_fee_basis_points
+                    } else {
+                        current_metadata.seller_fee_basis_points
+                    };
+
+                unprotected_upsert_asset_base_info(
```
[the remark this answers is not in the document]
danenbm: The idea is that all the "unprotected" upserts are blanket protected by the `update_metadata_seq`.

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
danenbm: I don't think it does anything for compressed assets and not sure why its there as it is updated by save_v1_asset.  Maybe this is an opportunity to just remove it from compressed indexing.

The reason it was turned into an "unprotected" upsert was just noting that it was to be blanket protected by `update_metadata_seq`.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: The naming `unprotected_upsert_creators` is actually a slight misnomer, as it actually had some protection inside it for the `verified` flag which you noticed and mentioned in later comment.

Example:
There's two creators, A and B.

`verify_creator` at `seq` 2 indexed first, and it inserts a row for creator B that only contains the creator address, `verified` set to true, `verifiedSeq` set to 2.

`mint_v1` at `seq` 1 is indexed next and inserts the row for creator A, but for creator B only updates the record for `postion`, `share`, and `slotUpdated`.   It leaves the `verified` flag as-is because of `verifiedSeq`.

If there happened to be an `update_metadata` afterwards that changed creator positions and verification status, it would have blocked both the `verify_creator` and the `mint_v1` due to the blanket protection from `update_metadata_seq`.

However, this all needs to be revisited anyways as you have shown that the blanket protection from `update_metadata_seq` is unsound approach with concurrent ingestor processes.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Misnomer.  I guess it should have been called "partially protected" or something.  The protection on the verified flag was needed for verify/unverify.  But the positions, share amounts, etc. would be blanket protected by `update_metadata_seq`.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: I agree the current approach is unsound with concurrent ingestor processes.  I forgot about this, as my docker testing only runs one ingester.  I felt there was something too easy about Version 2, but couldn't quite see it as I tested and thought about it, which is why I had kept Version 1 around.

The choices are:
1. Either go back to more of the [Version 1](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134) approach, or
2. look into furthering the use of Postgres transactions to resolve it.

After offline discussion with @linuskendall and @NicolasPennie we are planning on doing something more like Version 1, but trying to incorporate transactions to combine some of the upserts in the style of https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/142.  I just have to make sure I understand/implement the error handling correctly based on the comments from @linuskendall.

nft_ingester/src/program_transformers/bubblegum/delegate.rs
danenbm: The check didn't need to strictly be there.  The original version 1 for the `!asset_should_be_updated` check was actually just an `asset_was_decompressed` test.  The idea was to just say "if this thing has been decompressed, don't try to do any other compression indexing and just return immediately".

This version 1 check has the same problem with concurrency as pointed out in the other comment, so it won't be there in the next version.

digital_asset_types/src/dao/generated/asset_grouping.rs
```diff
@@ -12,15 +12,15 @@ impl EntityName for Entity {
...
+#[derive(Clone, Debug, PartialEq, DeriveModel, DeriveActiveModel, Serialize, Deserialize)]
 pub struct Model {
     pub id: i64,
     pub asset_id: Vec<u8>,
     pub group_key: String,
     pub group_value: Option<String>,
     pub seq: Option<i64>,
     pub slot_updated: Option<i64>,
-    pub verified: Option<bool>,
+    pub verified: bool,
```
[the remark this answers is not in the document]
danenbm: If you look at `m20230720_120101_add_asset_grouping_verified.rs` it was `not_null()` from that point.  All I did for this was regenerate the SeaORM objects based on the existing migrations.  Do we need a special migration to check whether people have it as Optional in their databases?  My guess is that it wasn't ever nullable and the SeaORM was somehow not updated?

digital_asset_types/src/dao/generated/cl_audits.rs
```diff
@@ -22,7 +22,7 @@ pub struct Model {
     pub seq: i64,
     pub level: i64,
     pub hash: Vec<u8>,
-    pub created_at: Option<DateTime>,
+    pub created_at: DateTime,
```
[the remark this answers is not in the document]
danenbm: Yeah I have same response.  If you look at `m20230919_072154_cl_audits.rs` its not nullable in the original migration that added the table.

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: This is part of the upserts that were not inherently protected by their own SQL logic, but blanket protected by `update_metadata_seq` that we now know has a concurrency issue.

nft_ingester/src/tasks/common/mod.rs
danenbm: >I can see that the download_metadata uri should be protected by a seq since that is on chain. But the offchain metadata can change at any time. To me the system should always download the latest metadata that is offchain irrespectively of the seq?

So the URI is already protected by its own sequence number in the `mint_V1` and `update_metadata` code, but the reason I added this seq is to make sure that a task doesn't update `asset_data.metadata` with downloaded data from an older task.

I think because URI can be changed, this can happen whether stuff is indexed in-order or out-of-order.

Example:
1. `mint_v1`, has uri1, seq 1, kicks off task(uri1, seq=1).
2. `update_metadata`, has uri2, seq2, kicks of task(uri2, seq=2)
3. Task2 finishes and asset_data.metadata to the data from uri2.
4. Task1 finishes and **incorrectly** sets `asset_data.metadata` to the data from uri1.

However, with using a download_metadata_seq, step 4 would be different.  Task1 would finish and NOT update `asset_data.metadata` to the data from uri1.

danenbm:
Due to some issues with this approach, most likely we are now going with [Version 1](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134)

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
danenbm: https://github.com/metaplex-foundation/blockbuster/pull/31

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: https://github.com/metaplex-foundation/blockbuster/pull/31

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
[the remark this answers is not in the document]
danenbm: These are all removed in: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
danenbm: Removed insert to `asset_v1_account_attachments` in Bubblegum parsing in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: Fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
danenbm: WIll incorporate in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134 after this PR is updated

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: Will incorporate in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134 after the above is merged.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
danenbm: Fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Went back to Version 1 approach: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134 and incorporated all learnings from Version 2 PR.

nft_ingester/src/program_transformers/bubblegum/delegate.rs
danenbm: Fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

nft_ingester/src/program_transformers/bubblegum/db.rs
danenbm: Fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134

migration/src/m20231019_120101_add_seq_numbers_bgum_update_metadata.rs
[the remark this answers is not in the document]
NicolasPennie: @linuskendall is right, it should be fine

danenbm:
Version 2 asset table checks can cause a race condition with concurrent ingestor processes.  Closing this PR in favor of [Version 1](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/134)

PR #137, opened 2023-11-06
Add GetSigntaturesForAsset API

nikhil-helius:
Add GetSigntaturesForAsset API


digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -339,3 +339,73 @@ pub async fn get_by_id(
...
+}
+
+pub async fn fetch_transactions(
+    conn: &impl ConnectionTrait,
+    tree: Vec<u8>,
+    leaf_id: i64,
+    pagination: &Pagination,
+    limit: u64,
+) -> Result<Vec<(String, Option<String>)>, DbErr> {
+    let mut stmt = cl_audits::Entity::find().filter(cl_audits::Column::Tree.eq(tree));
```
danenbm, line 395: Just overall thinking about the `cl_audits` table, which probably contains every tree that is indexed.  Won't this get pretty big and need to be purged sometimes like maybe after a tree is fully audited?  If that is the case do we want a public API method that depends on it?  Or should we just stick with stuff within the tools/ directory?

Also, another thing is that `cl_audits` table is configurable, so, for example, `insert_change_log`  takes a bool indicating whether to populate that table.  I think we need a similar check so that if `cl_items` is not present in the configuration, it returns some kind of "not available in this system" type of messsage.  I didn't see that anywhere, please correct me if I missed it.

danenbm (changes requested):
Overall code is nice and clean, but just had some systemic type of questions.

digital_asset_types/src/dao/scopes/asset.rs
NicolasPennie: I think that makes sense. The provider can decide whether they enable it or not. A user could also enable it for a subset of trees.

das_api/src/builder.rs
```diff
@@ -80,6 +80,12 @@ impl RpcApiBuilder {
             rpc_context.search_assets(payload).await.map_err(Into::into)
         })?;
         module.register_alias("searchAssets", "search_assets")?;
+        
+        module.register_async_method("get_signatures_for_asset", |rpc_params, rpc_context| async move {
+            let payload = rpc_params.parse::<GetSignaturesForAsset>()?;
+            rpc_context.get_signatures_for_asset(payload).await.map_err(Into::into)
+        })?;
+        module.register_alias("getSignaturesForAsset", "get_signatures_for_asset")?;
```
febo, line 88: Would it make sense to named this as `getAssetSignatures`? We have a `getAssetProof` so I wonder if this one fits the same pattern.

das_api/src/api/api_impl.rs
```diff
@@ -363,4 +364,34 @@ impl ApiContract for DasApi {
...
+    async fn get_signatures_for_asset(
+        self: &DasApi,
+        payload: GetSignaturesForAsset,
+    ) -> Result<TransactionSignatureList, DasApiError> {
+        let GetSignaturesForAsset { 
+            id,
+            limit,
+            page,
+            before,
+            after,
```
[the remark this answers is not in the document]
NicolasPennie: We could. We'd need a different keyset though, since it's based on the Seq and not the ID or createdAt field.

But page-based is reasonable for this API since the volume size isn't too large per key. It will iterate quickly enough over 1M records, and it shouldn't get much bigger than that.

das_api/src/builder.rs
NicolasPennie: We can add an alias in a second PR if people prefer. I wanted to keep it consistent with Solana RPC ("getSignaturesForAddress").

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -339,3 +339,73 @@ pub async fn get_by_id(
...
+        if tree.is_empty() {
+            return Err(DbErr::Custom("Empty tree for asset".to_string()));
+        }
+        let leaf_id = asset
+            .nonce
+            .ok_or(DbErr::RecordNotFound("Leaf ID does not exist".to_string()))?;
+        let transactions = fetch_transactions(conn, tree, leaf_id, pagination, limit).await?;
+        Ok(transactions)
+    } else {
+        Ok(Vec::new())
```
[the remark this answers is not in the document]
NicolasPennie: We can fix. When you query by leafId + tree empty array makes more sense in my opinion. Either work though.

danenbm:
We can close this PR now right as its covered by https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/155?

danenbm:
Closing this PR in favor of https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/155

PR #138, opened 2023-11-08
Proposal for AGPL licensing

linuskendall:
_@danenbm UPDATE:  I am currently working on adding all contributors who need to approve the license change to this PR.  Some are still in the process of being invited as a collaborator so that I can list them as a reviewer.  We need to wait to merge this until all needed contributors approve, or we have offline sign off._

This repo needs a license, and the more contributors we have to the repo the greater urgency to actually define that license. Each contributor will have to agree/sign off on the license we pick.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
+
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+our General Public Licenses are intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
+software for all its users.
+
+  When we speak of free software, we are referring to freedom, not
+price.  Our General Public Licenses are designed to make sure that you
+have the freedom to distribute copies of free software (and charge for
+them if you wish), that you receive source code or can get it if you
+want it, that you can change the software or use pieces of it in new
+free programs, and that you know you can do these things.
+
+  Developers that use our General Public Licenses protect your rights
+with two steps: (1) assert copyright on the software, and (2) offer
+you this License which gives you legal permission to copy, distribute
+and/or modify the software.
+
+  A secondary benefit of defending all users' freedom is that
+improvements made in alternate versions of the program, if they
+receive widespread use, become available for other developers to
+incorporate.  Many developers of free software are heartened and
+encouraged by the resulting cooperation.  However, in the case of
+software used on network servers, this result may fail to come about.
+The GNU General Public License permits making a modified version and
+letting the public access it on a server without ever releasing its
+source code to the public.
+
+  The GNU Affero General Public License is designed specifically to
+ensure that, in such cases, the modified source code becomes available
+to the community.  It requires the operator of a network server to
+provide the source code of the modified version running there to the
+users of that server.  Therefore, public use of a modified version, on
+a publicly accessible server, gives the public access to the source
+code of the modified version.
+
+  An older license, called the Affero General Public License and
+published by Affero, was designed to accomplish similar goals.  This is
+a different license, not a version of the Affero GPL, but Affero has
+released a new version of the Affero GPL which permits relicensing under
+this license.
+
+  The precise terms and conditions for copying, distribution and
+modification follow.
+
+                       TERMS AND CONDITIONS
+
+  0. Definitions.
+
+  "This License" refers to version 3 of the GNU Affero General Public License.
+
+  "Copyright" also means copyright-like laws that apply to other kinds of
+works, such as semiconductor masks.
+
+  "The Program" refers to any copyrightable work licensed under this
+License.  Each licensee is addressed as "you".  "Licensees" and
+"recipients" may be individuals or organizations.
+
+  To "modify" a work means to copy from or adapt all or part of the work
+in a fashion requiring copyright permission, other than the making of an
+exact copy.  The resulting work is called a "modified version" of the
+earlier work or a work "based on" the earlier work.
+
+  A "covered work" means either the unmodified Program or a work based
+on the Program.
+
+  To "propagate" a work means to do anything with it that, without
+permission, would make you directly or secondarily liable for
+infringement under applicable copyright law, except executing it on a
+computer or modifying a private copy.  Propagation includes copying,
+distribution (with or without modification), making available to the
+public, and in some countries other activities as well.
+
+  To "convey" a work means any kind of propagation that enables other
+parties to make or receive copies.  Mere interaction with a user through
+a computer network, with no transfer of a copy, is not conveying.
+
+  An interactive user interface displays "Appropriate Legal Notices"
+to the extent that it includes a convenient and prominently visible
+feature that (1) displays an appropriate copyright notice, and (2)
+tells the user that there is no warranty for the work (except to the
+extent that warranties are provided), that licensees may convey the
+work under this License, and how to view a copy of this License.  If
+the interface presents a list of user commands or options, such as a
+menu, a prominent item in the list meets this criterion.
+
+  1. Source Code.
+
+  The "source code" for a work means the preferred form of the work
+for making modifications to it.  "Object code" means any non-source
+form of a work.
+
+  A "Standard Interface" means an interface that either is an official
+standard defined by a recognized standards body, or, in the case of
+interfaces specified for a particular programming language, one that
+is widely used among developers working in that language.
+
+  The "System Libraries" of an executable work include anything, other
+than the work as a whole, that (a) is included in the normal form of
+packaging a Major Component, but which is not part of that Major
+Component, and (b) serves only to enable use of the work with that
+Major Component, or to implement a Standard Interface for which an
+implementation is available to the public in source code form.  A
+"Major Component", in this context, means a major essential component
+(kernel, window system, and so on) of the specific operating system
+(if any) on which the executable work runs, or a compiler used to
+produce the work, or an object code interpreter used to run it.
+
+  The "Corresponding Source" for a work in object code form means all
+the source code needed to generate, install, and (for an executable
+work) run the object code and to modify the work, including scripts to
+control those activities.  However, it does not include the work's
+System Libraries, or general-purpose tools or generally available free
+programs which are used unmodified in performing those activities but
+which are not part of the work.  For example, Corresponding Source
+includes interface definition files associated with source files for
+the work, and the source code for shared libraries and dynamically
+linked subprograms that the work is specifically designed to require,
+such as by intimate data communication or control flow between those
+subprograms and other parts of the work.
+
+  The Corresponding Source need not include anything that users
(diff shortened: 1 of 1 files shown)


fanatid (approved):

danenbm (approved):

NicolasPennie:
What do people think about using MIT instead? It's simpler and ubiquitous.

Also, given that this repository is meant to be a reference implementation, it seems logical that it should use a permissive license.

NicolasPennie:
> I think as a reference implementation its even more important that if people use components from this that those changes are contributed upstream, hence I think AGPL is the right license for this type of project.

By definition, a reference implementation suits a permissive license because then it allows for people to truly use it as a reference – for anything!

> Licensing as MIT would just make it much more likely that the reference implementation does not get upstream contributions from teams that run it, and therefore would not be a valuable asset for the Solana ecosystem as a whole. 

I don't agree. AGPL is controversial license in the software ecosystem for a few reasons, but I'll try to reduce the discussion to the key points that impact this repository.

Contributions come from aligning incentives. It might be counter-intuitive but permissive licenses (MIT, Apache, etc) tend to maximize adoption, and that leads to more contributions overall. Solana users want a consistent experience across their API and between vendors. That's the incentive for upstreaming changes and it's why Helius' policy is to immediately upstream any bug-fix we find for this implementation.

Less in scope at the moment, but It's worth noting that some organizations even ban the use of AGPL software. That might also impact adoption.

> I think for the plugin the right license is Apache, but for the RPC server and the ingester the license of AGPL makes the most sense.

Interesting. I think AGPL is much more fitting for the blockbuster/plugin than for this repository. The reason is that they are well-scoped and standalone solutions.

tsmbl (approved):

NicolasPennie:
I agree having a consistent DAS API is important – but we shouldn’t mix that up into the AGPL argument.

AGPL does not enforce collaboration, nor does it require parties to upstream their changes. The incentive for providers to maintain a consistent, quality reference is independent from the license. Helius has consistently upstreamed critical (and minor) bug fixes since launch, and we have added new features/endpoints.

The best thing for the ecosystem is to accelerate and promote innovation — it feels suboptimal to introduce friction in iteration speed for something as early as DAS. Instead, we want to maximize flexibility. For example, with an MIT/Apache license, teams can immediately leverage the existing build/deployment tooling within their private repositories. Teams are incentivized to experiment with their customers & build new features. Teams can also build custom modifications for specific customers that may not be suitable for a broader audience (e.g. query parameter that only works for a subset of records). Nothing stops other parties from adding those extensions to the core feature set if they gain traction. We’ve done this with GetAssetBatch & GetAssetProofBatch, and will continue doing so with other features once we are sufficiently comfortable with their quality, design, and adoption.

My fear is that we end up with something similar to the existing RPC API. Yes, it’s consistent. But the dev experience is poor and we haven’t seen anyone contribute improvements – ever. And now we have to rework it with RPC 2.0, where discussions have been stale and slow moving for many months.

linuskendall:
> I agree having a consistent DAS API is important – but we shouldn’t mix that up into the AGPL argument.
> 
> AGPL does not enforce collaboration, nor does it require parties to upstream their changes. The incentive for providers to maintain a consistent, quality reference is independent from the license. Helius has consistently upstreamed critical (and minor) bug fixes since launch, and we have added new features/endpoints.

It does certainly enforce that the changes are open source. If people then want to spend the effort of upstreaming or not that's another issue - one would hope that people would show that leadership to upstream valuable work they build on top of things but even they don't if it's open source others can do that work for them!

> The best thing for the ecosystem is to accelerate and promote innovation — it feels suboptimal to introduce friction in iteration speed for something as early as DAS. Instead, we want to maximize flexibility. For example, with an MIT/Apache license, teams can immediately leverage the existing build/deployment tooling within their private repositories. Teams are incentivized to experiment with their customers & build new features. Teams can also build custom modifications for specific customers that may not be suitable for a broader audience (e.g. query parameter that only works for a subset of records). Nothing stops other parties from adding those extensions to the core feature set if they gain traction. We’ve done this with GetAssetBatch & GetAssetProofBatch, and will continue doing so with other features once we are sufficiently comfortable with their quality, design, and adoption.

There's no problem keeping a private repo if you also keep a public one. You can use all your CI tooling that you want and yes you can include customer specific extensions into it as long as they are open sourced. Doesn't have to be that it is easily upstreamable but if you face this issue then chances are someone else faces a similar issue.

AGPL doesn't as you say, force you to upstream anything. The only thing it adds is a requirement to make the source available for the things that you build. That can be in very rough format or it can be in production ready format. 

> 
> My fear is that we end up with something similar to the existing RPC API. Yes, it’s consistent. But the dev experience is poor and we haven’t seen anyone contribute improvements – ever. And now we have to rework it with RPC 2.0, where discussions have been stale and slow moving for many months.

Well.. as one of the top providers in the ecosystem - if the RPC 2.0 work is slow then maybe it's time to step up and show leadership not just for your customers but for the ecosystem as a whole? Rather than complaining that "we haven't seen anyone ever contribute anything". 

Really - I think the plethora of geyser plugins, Old Faithful and Yellowstone gRPC/dragon's mouth are certainly contributions to make RPC2.0 happen. These things have been contributed to by different parties - Blockdaemon, Genesys back in the day and Triton. Holaplex open sourced their stack for NFT based indexing and it's available for anyone to adopt.

That's certainly the approach that we are taking and trying to open source all the components we've built for RPC2. 

I will also remind you that the Solana RPC stack is apache licensed. So your argument that MIT/apache will somehow spur innovation more than AGPL seems to be still not supported by the evidence - either here in the work on DAS API nor in the RPC stack.

linuskendall:
> Hmm if we were to factor out the api from the indexing code that allows closed source use of the indexing aide while enforcing that api users create open source clone and open source their changes down the line.
> 
> A crate that provides a solid interface for queries to a data source to fill in the api responses would be a part of that, then in your api implementation you are not forced to open source your indexing code or your database accessor code. You only are open sourcing your modifications to the api and the datasource interface.
> 
> Thoughts?
> 
> Then the indexing code here would go to a new repository under mit license but the api spec, rust code and data source interface remain here in a license that ensures modifications must be open source.

I have had some thoughts on this :)

In practice: AGPL does not have anything that can be enforced for the indexing code (nft_ingesters, program_transformers) since these run internally and are never user facing ("not served over a network" in the parlance of AGPL). This means that even if we license the whole repo AGPL people can **still** run the nft_ingester and program transformers using private forks/modifications without publishing their source code since this would fall under "internal use". 

I still think AGPL would be the best license as it signal intention - however I'm not going to get stuck on that if it doesn't have real world, practical implications. In that case I think it would be better to license parts of this repo with a permissive license like Apache if that's what the contributors are can agree on / are willing to accept.  

AGPL provides one additional benefit over Apache or MIT. If we license program_transformers and ingester as Apache/MIT then in theory someone can distribute these as part of a closed source package. I think this is unlikely to happen but sure it's possible. 

I agree with Austin that no matter what the frontend code really needs to be licensed AGPL. This is the best way in my mind that we can ensure interoperability and sharing of modifications to the API responses. Yes, it doesn't force anyone to upstream but at least it forces everyone to publish their changes somewhere so if one day someone says that Triton has modifications X,Y or Z then anyone else can come and find the source for those modifications and pull them from wherever we have published them.

We do not need to split the repo to license different parts with different licenses. We just need to drop LICENSE files in the separate folders and then have an overall explanation of the licensing. This is done in multiple other cases.

NicolasPennie:
> I agree with Austin that no matter what the frontend code really needs to be licensed AGPL. This is the best way in my mind that we can ensure interoperability and sharing of modifications to the API responses. Yes, it doesn't force anyone to upstream but at least it forces everyone to publish their changes somewhere so if one day someone says that Triton has modifications X,Y or Z then anyone else can come and find the source for those modifications and pull them from wherever we have published them.

> We do not need to split the repo to license different parts with different licenses. We just need to drop LICENSE files in the separate folders and then have an overall explanation of the licensing. This is done in multiple other cases.

I'm aligned licensing the API with AGPL and the remaining code as MIT/Apache.

linuskendall:
> hooray we have a conclusion.

Not so fast there cowboy. I've been made aware that both blockbuster and plerkle-messenger is licensed under AGPL. Since they are both used as libraries by the ingester this means in my IANAL understanding that this repo absolutely needs to be fully under AGPL.

However, I also don't think that this stops anyone from running a private fork of the ingester for internal purposes assuming that:

- they are not distributing it in in like a compiled version (they would need to distribute the source then)
- they are not serving it over the network (they are not - that's the das_api frontend component, which we have anyway agreed can remain AGPL).

Therefore I'm back to my original (and continuing) stance that AGPL is the right license for this repo :)

linuskendall:
> Lets change the license for those, and put a nail in this. I'm the idiot who made those AGPL probbably.

The licensing for those would be owned by Metaplex, so it would be up to Metaplex (and any external contributors) to change the license if this was desired.

I am still not clear what the exact objections to AGPL for the indexer is. We already agree that AGPL should apply to the api and frontend. From what I can tell, AGPL for the indexer and libraries really doesn't make a practical difference for those who want to run private forks. The only limitation it places is if someone wants to compile a private version of the ingester and sell it as a compiled binary? 

It just doesnt seem like this is a likely scenario and that AGPL wouldn't really place restrictions here. FWIW we would probably look to license the code that @fanatid is contributing for the new gRPC based indexer as AGPL.

MetaplexFoundation:
Thank you all for the spirited discussion and feedback. Upon further review, the fact that this repo is not already licensed under AGPL 3.0 is an error. This repo added code from Blockbuster in September 2022 as a modification to implement account and data parsing. Blockbuster has been licensed under AGPL since August 2022.

Because this repo: (i) imports and runs the Blockbuster code; and (ii) has been conveyed publicly such that others can make copies, it must be licensed under AGPL 3.0 (see section 5).

However, as noted by linuskendall, the application of AGPL 3.0 to this repo does not necessarily mean that modified versions of code from this repo must be open sourced. If the modified code is run internally and not accessed by users over a network, it can remain close sourced.

Given the foregoing, we are going to update this repo to reflect the correct AGPL 3.0 license.

linuskendall:
> 
> Given the foregoing, we are going to update this repo to reflect the correct AGPL 3.0 license.
>

As is obvious in this thread, I generally agree with @MetaplexFoundation here  that this needs to be AGPL for all the reasons mentioned above and more. However, I will also say that this repo does not come with a CLA/copyright assignment nor did it have a specific license (mistake or otherwise) before this (hence why I brought up this issue in private many many moons ago before pushing this ticket).

This means that everyone who contributed (or more likely - the  organisations they worked for at the time i.e. Metaplex, Triton and Helius primarily - maybe I missed someone else?) retains copyright over their contributions and have not (as yet) licensed them to anyone under any license terms. Metaplex would not be able to single handedly decided the license for code that others contributed and has copyright over. 

We could argue that there is an implicit license due to the dependency on other AGPL licensed software, but this is still quite dicey since there wasn't an explicit license at the time of contriubtions.

Triton and any contributions made by people working in our team are all signed up for AGPL. However, significant (and excellent) portions of code have been contributed by @NicolasPennie, @nikhil-helius and @tahsintunan, so Helius would also need to agree with the licensing for this to be applied. The other option would be to exclude code from any contributor that didn't agree with the licensing, which would be a bad outcome that I'm sure nobody wants.

enzotar:
MPL 2.0 would satisfy both sides

NicolasPennie (approved):
We firmly believe that a permissive license (Apache/MIT) would be better suited for this repository (outlined in earlier comments). We do not support the more restrictive and opaque AGPL license.

We believe that _some_ open-source license is better than no license. And for that reason, **we reluctantly agree to this licensing proposal.**

We strongly encourage that everyone reconsiders and opts for an Apache license instead.

enzotar:
Why isn't MPL 2.0 considered?

PR #139, opened 2023-11-09
Update raw name and raw symbol for existing NFTs

linuskendall:
Update the raw name and raw symbol from the onchain data for existing NFTs when reingesting. This allows correction of incorrect values during reprocessing on an existing index.


danenbm:
Note I do have both of these fixed in https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/135 so we could approve this but also could approve with update metadata deployment

danenbm (approved):
Approved but note we will automatically get this when https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/135 is approved as well.

PR #140, opened 2023-11-20
ingester: fix ctrl-c handler

fanatid:
The current Ctrl-C handler doesn't make any sense, everything works the same even if we do not have this handler.
Instead of returning `JoinHandle` from the functions I made them async and used `try_join_all` with signals (sigterm ,sigint). Right now all tasks actually never return `Err` but we can change it in the future and signals handler still would work correctly.

fmt, clippy (+rust update) can be moved to another PR, if required


fanatid:
Need to updated after #141

PR #144, opened 2023-11-24
Remove program_transformers dependency on plerkle

fanatid:
## Goal
Enable the development of alternative ingestion engines and event sources for DAS.

## How
Building on a prior PR that isolated program_transforms into a separate crate for managing business logic around account categorization, decoding, and saving, this PR extends the generalization of program_transforms. It transitions from using plerkle to native Solana structs for accounts and transactions.

## Why
With program_transforms no longer dependent on plerkle, Triton can deploy a new ingester version utilizing dragonmoth, which leverages protobufs for messaging.

## Changes
- Convert program_transforms to be based on solana structs.
- Use parsers from plerkle_serialization to transform the plerkle structs into solana struct in the ingester.
- A recent update of blockbuster updates to solana 1.17 so doing the same here. anchor updated to 0.29.

## Release Steps
There are 2 dependent PRs that must be merged and crate versions released before this PR.
- 1. [x] https://github.com/metaplex-foundation/digital-asset-validator-plugin/pull/70
- 2. [x] https://github.com/metaplex-foundation/blockbuster/pull/38

 #143


fanatid:
Updated. Required blockbuster update: https://github.com/metaplex-foundation/blockbuster/pull/38

WilfredAlmeida:
Thanks for this separation. We are now using the separated `program_transformers` in https://github.com/WilfredAlmeida/LightDAS

nft_ingester/src/plerkle.rs
```diff
@@ -0,0 +1,74 @@
...
+                .ok_or(PlerkleDeserializerError::NotFound)?,
+        )
+        .try_into()?;
+        let compiled = tx_info.compiled_inner_instructions();
+        let inner = tx_info.inner_instructions();
+        let meta_inner_instructions = compiled
+            .map(|c| PlerkleCompiledInnerInstructionVector(c).try_into())
+            .or_else(|| inner.map(|i| PlerkleInnerInstructionsVector(i).try_into()))
+            .transpose()?
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
fanatid: hm, no blockbuster looks different https://github.com/metaplex-foundation/blockbuster/blob/main/blockbuster/src/instruction.rs#L63-L100

something like this?
```
let meta_inner = match compiled {
  Some(c) => PlerkleCompiledInnerInstructionVector(c).try_into(),
  None => PlerkleInnerInstructionsVector(inner.unwrap()).try_into(),
}?;
```

program_transformers/src/lib.rs
```diff
@@ -28,6 +28,23 @@ pub mod error;
 mod token;
 mod token_metadata;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AccountInfo {
+    pub slot: u64,
+    pub pubkey: Pubkey,
+    pub owner: Pubkey,
+    pub data: Vec<u8>,
+}
```
danenbm, line 37: Should we keep `write_version` in here from the original serialized types?  Maybe this is a better discussion for https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/issues/177 but I have been suspecting we might want to utilize it at some point in the future.

Background on what that value is: https://docs.rs/solana-geyser-plugin-interface/1.18.5/solana_geyser_plugin_interface/geyser_plugin_interface/struct.ReplicaAccountInfoV2.html#structfield.write_version

It might help us fix something that could be affecting us in account-based updates.

fanatid: Depending on the commitment level that we want to support, `write_version` is only important for processed. I'd recommend staying only on finalized, support processed + confirmed is a huge pain from our experience in the steamboat (custom gPA indices).

danenbm: ok, yeah I did not know it only would be used for processed and confirmed.  I assumed if there were two different account updates in the same slot in different transactions, it would be used even at finalized.  But if not, we might as well leave it out.

PR #148, opened 2023-11-28
Configurable account streams

linuskendall:
This makes the worker configuration more dynamic, by letting you specify custom streams in the config file. It also removes an unnecessary dependency between plerkle_messenger and the ingester in pulling the names of the stream from plerkle messenger.

The default behaviour of the ingester stays the same, creating 2 workers for TXN and 2 workers for ACC streams by default.

### DIFF
--- a/nft_ingester/src/config.rs
+++ b/nft_ingester/src/config.rs
@@ -27,13 +27,23 @@ pub struct IngesterConfig {
     pub backfiller_trees: Option<Vec<String>>,
     pub role: Option<IngesterRole>,
     pub max_postgres_connections: Option<u32>,
-    pub account_stream_worker_count: Option<u32>,
-    pub account_backfill_stream_worker_count: Option<u32>,
-    pub transaction_stream_worker_count: Option<u32>,
-    pub transaction_backfill_stream_worker_count: Option<u32>,
+    pub worker_config: Option<Vec<WorkerConfig>>,
     pub code_version: Option<&'static str>,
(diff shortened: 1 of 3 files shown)


nft_ingester/src/config.rs
```diff
@@ -66,20 +76,31 @@ impl IngesterConfig {
...
+                    stream_name: "ACC".to_string(),
+                    worker_count: 2,
+                    worker_type: WorkerType::Account,
+                },
+                WorkerConfig {
+                    stream_name: "TXN".to_string(),
+                    worker_count: 2,
+                    worker_type: WorkerType::Transaction,
+                },
+            ]
```
kespinola, line 94: If worker configs are tied to the stream shouldn't they be derived from the worker type to avoid misconfiguration?

Also related, what are folks thoughts on using clap for this project and having different command entry points for the different worker types instead of using an environment variable. Idea is to provide more structure to the code base and promote providers running the workers using separate processes as suggested in the production guide.

fanatid: > different command entry points for the different worker types instead of using an environment variable

I hope that we will test and migrate to the new ingester soon :slightly_smiling_face:  it's possible to define streams to consume in config with a lot of options https://github.com/rpcpool/digital-asset-rpc-infrastructure/blob/pt-ingester2/nft_ingester2/config-ingester.yml

danenbm (approved):

PR #155, opened 2024-01-10
feat: `getSignaturesForAsset` endpoint on top of new `cl_audits_v2` table

tahsintunan:
# Overview:
- New `cl_audits_v2` table.
- `getAssetSignatures` endpoint
- Add `update_metadata` instruction via migration.


nft_ingester/src/transaction_notifications.rs
```diff
@@ -22,13 +22,12 @@ pub fn transaction_worker<T: Messenger>(
...
     ack_channel: UnboundedSender<(&'static str, String)>,
     consumption_type: ConsumptionType,
-    cl_audits: bool,
     stream_key: &'static str,
 ) -> JoinHandle<()> {
     tokio::spawn(async move {
         let source = T::new(config).await;
         if let Ok(mut msg) = source {
-            let manager = Arc::new(ProgramTransformer::new(pool, bg_task_sender, cl_audits));
+            let manager = Arc::new(ProgramTransformer::new(pool, bg_task_sender));
```
NicolasPennie, line 30: Don't remove the cl_audits flag. It's configurable on this version.

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -97,46 +87,39 @@ where
...
             .await
             .map_err(|db_err| IngesterError::StorageWriteError(db_err.to_string()))?;
-
-        // Insert the audit item after the insert into cl_items have been completed
-        if let Some(audit_item) = audit_item {
-            cl_audits::Entity::insert(audit_item).exec(txn).await?;
-        }
     }
 
-    // If and only if the entire path of nodes was inserted into the `cl_items` table, then insert
```
NicolasPennie, line 107: Don't remove this, this is used for the backfiller

das_api/src/builder.rs
```diff
@@ -94,6 +94,17 @@ impl RpcApiBuilder {
         )?;
         module.register_alias("getAssetsByGroup", "get_assets_by_group")?;
 
+        module.register_async_method(
```
NicolasPennie, line 97: I believe the other maintainers wanted an alias to be added? @danenbm do you still need this?

NicolasPennie (changes requested):

das_api/src/builder.rs
danenbm: I think other maintainers did strongly advocate for `getAssetProofs`, so it does make sense to add `getAssetSignatures` to match that API format that we probably want to standardize.

Even though I understand that `getSignaturesForAsset` follows a Solana RPC ("getSignaturesForAddress") format.

NicolasPennie: sure, adding an alias doesn't matter much to me

digital_asset_types/src/dao/generated/sea_orm_active_enums.rs
```diff
@@ -123,3 +123,80 @@ pub enum V1AccountAttachments {
...
+            "Delegate" => Instruction::Delegate,
+            "MintToCollectionV1" => Instruction::MintToCollectionV1,
+            "MintV1" => Instruction::MintV1,
+            "Redeem" => Instruction::Redeem,
+            "SetAndVerifyCollection" => Instruction::SetAndVerifyCollection,
+            "Transfer" => Instruction::Transfer,
+            "UnverifyCollection" => Instruction::UnverifyCollection,
+            "UnverifyCreator" => Instruction::UnverifyCreator,
+            "VerifyCollection" => Instruction::VerifyCollection,
+            "VerifyCreator" => Instruction::VerifyCreator,
```
NicolasPennie, line 178: Missing UpdateMetadata

NicolasPennie (approved):

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -427,6 +429,83 @@ pub async fn get_by_id(
...
+        ));
+    }
+
+    // if only asset_id is provided, fetch the latest tree and leaf_idx (asset.nonce) for the asset
+    // and use them to fetch transactions
+    let stmt = asset::Entity::find()
+        .distinct_on([(asset::Entity, asset::Column::Id)])
+        .filter(asset::Column::Id.eq(asset_id))
+        .order_by(asset::Column::Id, Order::Desc)
+        .limit(1);
```
kespinola, line 490: getSignaturesForAddress provided by vanilla solana PRC orders the transactions by desc slot. If its desired to mimic that call but for compression trees then should order by seq.

kespinola: index on seq should be added if order is switched.

NicolasPennie: This is fetching the asset to lookup the tree and leaf info. You actually don't even need the `order_by` since asset_id is a unique field.

NicolasPennie: You'll see that `fetch_transactions` orders by Seq ascending. So the opposite direction of `getSignauresForAddress`. 

In hindsight, doing DESC would have been better. Ideally I'd rather not change it now to avoid breaking our customers. If reviewers feel strongly about it we can discuss with our customers, too.

danenbm: If its a single or minimal amount of customers who are impacted by the ordering, might be worth asking them quick if they depend on the order and if it would break their use case.

I think ideally we would match `getSignaturesForAddress` if there's no or very minimal impact.  I agree if it breaks them or is some work, we don't want to change it.  But if you could just ask them and they say it doesn't matter then we should change it imo.

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -97,10 +92,37 @@ where
...
+    if cl_audits {
+        let tx_id_bytes = bs58::decode(txn_id)
+            .into_vec()
+            .map_err(|_e| IngesterError::ChangeLogEventMalformed)?;
+        let audit_item_v2 = cl_audits_v2::ActiveModel {
+            tree: Set(tree_id.to_vec()),
+            leaf_idx: Set(change_log_event.index as i64),
+            seq: Set(change_log_event.seq as i64),
+            tx: Set(tx_id_bytes),
+            instruction: Set(Instruction::from_str(instruction)),
```
danenbm, line 107: Should this log a metric/error if the `Instruction::from_str` conversion results in `Instruction::Unknown`?

digital_asset_types/src/dao/generated/mod.rs
```diff
@@ -10,6 +10,7 @@ pub mod asset_grouping;
 pub mod asset_v1_account_attachments;
 pub mod backfill_items;
 pub mod cl_audits;
```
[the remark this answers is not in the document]
NicolasPennie: With the current way the ORM works this is not possible. However I untangled the entire ORM circular dependency and I will upstream that soon. 

So TLDR, it will be fixed in a subsequent PR!

migration/src/lib.rs
```diff
@@ -71,6 +74,9 @@ impl MigratorTrait for Migrator {
             Box::new(m20230919_072154_cl_audits::Migration),
             Box::new(m20231019_120101_add_seq_numbers_bgum_update_metadata::Migration),
             Box::new(m20231206_120101_remove_was_decompressed::Migration),
+            Box::new(m20240104_203133_add_cl_audits_v2::Migration),
+            Box::new(m20240104_203328_remove_cl_audits::Migration),
```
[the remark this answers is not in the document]
NicolasPennie: The reason to use a different table is so that providers (like ourselves) can make the change with zero downtime. You can run both in parallel if you need to.

Note that the new table stores all data as "bytea" and uses an enum to reduce storage. These changes reduced our table size from about 800GB to 46GB!

nft_ingester/src/program_transformers/bubblegum/db.rs
NicolasPennie: Sure

NicolasPennie: Not too critical from my perspective given that the instructions are known and we are well-aware of all changes. But defensiveness is good.

migration/src/lib.rs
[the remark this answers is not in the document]
danenbm: Ah I did not notice that but that is significant!!!

digital_asset_types/src/dao/scopes/asset.rs
danenbm: To reiterate this comment is not a blocker just a request if its quick to do.

danenbm (approved):
LGTM!

PR #156, opened 2024-01-11
fix: remove decompress ix handling and use db transactions

NicolasPennie:
## Overview

* Remove decompress ix handling for cNFTs. It's redundant. Instead we should fully rely on the indexing of the regular NFT via accounts (new source of truth)
*  Fix out-of-order issue between txns and accounts. When indexing the metadata account, we allow the update for an asset if `slot_updated` is null. This happens when there is already a corresponding row in the asset table from cNFT events.
*  Properly wipe the data/creator hash to enforce consistency for decompressed NFTs.
*  Group all DB operations for a single cNFT ix within one DB transaction. Cleaner to read / maintain, but also marginally faster on the DB and enforces consistent reads for customers. Without a DB transaction a customer might see a newly minted cNFT without a collection.  

## Testing
Running in Helius prod


nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -106,11 +111,6 @@ where
...
                 };
 
-                // Begin a transaction.  If the transaction goes out of scope (i.e. one of the executions has
-                // an error and this function returns it using the `?` operator), then the transaction is
-                // automatically rolled back.
-                let multi_txn = txn.begin().await?;
-
                 // Upsert `asset` table base info and `asset_creators` table.
                 upsert_asset_base_info(
                     txn,
```
danenbm, line 116: This is an existing bug, should have been `multi_txn`.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -163,8 +163,6 @@ where
 
                 upsert_asset_with_seq(&multi_txn, id_bytes.to_vec(), seq as i64).await?;
 
-                multi_txn.commit().await?;
-
                 // Upsert creators to `asset_creators` table.
                 upsert_asset_creators(
                     txn,
```
danenbm, line 168: I think your intention would be for this one to be `multi_txn` as well.

nft_ingester/src/program_transformers/bubblegum/mint_v1.rs
```diff
@@ -204,6 +202,8 @@ where
                     return Ok(None);
                 }
 
+                multi_txn.commit().await?;
+
```
danenbm, line 206: This needs to be above the URI check or it will return before committing all the other info.

nft_ingester/src/program_transformers/bubblegum/update_metadata.rs
```diff
@@ -147,7 +152,7 @@ where
                 let multi_txn = txn.begin().await?;
```
danenbm, line 152: This is now a duplicate and should be removed.

nft_ingester/src/program_transformers/token_metadata/v1_asset.rs
```diff
@@ -223,6 +223,8 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
                     asset::Column::Leaf,
                     asset::Column::Compressed,
                     asset::Column::Compressible,
+                    asset::Column::DataHash,
+                    asset::Column::CreatorHash,
```
danenbm, line 227: Good catch on missed columns.

Nitpick: I would put these in same order as they are in the model above (hashes first then compressed/compressible) just for consistency.

danenbm (changes requested):

nft_ingester/src/program_transformers/token_metadata/v1_asset.rs
```diff
@@ -234,7 +236,7 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
         )
         .build(DbBackend::Postgres);
     query.sql = format!(
-        "{} WHERE excluded.slot_updated > asset.slot_updated",
+        "{} WHERE excluded.slot_updated > asset.slot_updated OR asset.slot_updated IS NULL",
```
danenbm, line 239: Good catch here too!

danenbm (approved):

PR #157, opened 2024-01-16
Improvements for NFT indexing and query consistency

NicolasPennie:
## Overview
* Indexing improvements for regular NFTs.
    * Fix parsing for ownership types based on token standard. 
    *  If a token standard is unknown, but supply=1, then mark it as an NFT. If supply > 1, mark it as fungible.
    * In v1_asset, retry the fetch for a mint/token account since all accounts can arrive at once during an NFT mint.
 * Query only well-formed tokens by default w/ search assets (like getAssetsByOwner). Only query NFTs, since only NFTs are supported at the moment. These changes avoid inconsistencies in the responses that confused our customers.
    * TLDR: Your queries only include well-defined NFTs unless you configure flags like supply/owner_type yourself (no one really does).

## Testing
These changes are running in Helius production, but I also ran the code locally and spot checked against regular NFTs.

### DIFF
--- a/digital_asset_types/src/dao/mod.rs
+++ b/digital_asset_types/src/dao/mod.rs
@@ -73,78 +73,12 @@ pub struct SearchAssetsQuery {
 }
 
 impl SearchAssetsQuery {
-    pub const fn count_conditions(&self) -> usize {
-        // Initialize counter
-        let mut num_conditions = 0;
-        if self.specification_version.is_some() {
-            num_conditions += 1;
-        }
-        if self.specification_asset_class.is_some() {
-            num_conditions += 1;
-        }
-        if self.owner_address.is_some() {
-            num_conditions += 1;
-        }
-        if self.owner_type.is_some() {
-            num_conditions += 1;
-        }
-        if self.delegate.is_some() {
-            num_conditions += 1;
-        }
-        if self.frozen.is_some() {
-            num_conditions += 1;
-        }
-        if self.supply.is_some() {
-            num_conditions += 1;
-        }
-        if self.supply_mint.is_some() {
-            num_conditions += 1;
-        }
-        if self.compressed.is_some() {
-            num_conditions += 1;
-        }
-        if self.compressible.is_some() {
-            num_conditions += 1;
-        }
-        if self.royalty_target_type.is_some() {
-            num_conditions += 1;
-        }
-        if self.royalty_target.is_some() {
-            num_conditions += 1;
-        }
-        if self.royalty_amount.is_some() {
-            num_conditions += 1;
-        }
-        if self.burnt.is_some() {
-            num_conditions += 1;
-        }
-        if self.creator_address.is_some() {
-            num_conditions += 1;
-        }
-        if self.creator_address.is_some() {
-            num_conditions += 1;
-        }
-        if self.grouping.is_some() {
-            num_conditions += 1;
-        }
-        if self.json_uri.is_some() {
-            num_conditions += 1;
-        }
-        if self.name.is_some() {
-            num_conditions += 1;
-        }
-
-        num_conditions
-    }
-
     pub fn conditions(&self) -> Result<(Condition, Vec<RelationDef>), DbErr> {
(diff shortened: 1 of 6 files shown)


digital_asset_types/src/dao/mod.rs
```diff
@@ -197,6 +125,36 @@ impl SearchAssetsQuery {
...
+        } else {
+            // Default to NFTs
+            //
+            // In theory, the owner_type=single check should be sufficient,
+            // however there is an old bug that has marked some non-NFTs as "single" with supply > 1.
+            // The supply check guarentees we do not include those.
+            conditions = conditions.add_option(Some(
+                asset::Column::OwnerType
+                    .eq(OwnerType::Single)
+                    .and(asset::Column::Supply.lte(1)),
```
kespinola, line 154: If trees and accounts are backfilled will they be corrected?

Related patch for mis-labled owner type for non-NFTs
 https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/151

digital_asset_types/src/rpc/filter.rs
```diff
@@ -11,7 +11,7 @@ pub struct AssetSorting {
 impl Default for AssetSorting {
     fn default() -> AssetSorting {
         AssetSorting {
-            sort_by: AssetSortBy::Created,
+            sort_by: AssetSortBy::Id,
```
kespinola, line 14: This may affect callers of the API because expecting created sort and switched to id.

kespinola (approved):

digital_asset_types/src/rpc/filter.rs
NicolasPennie: Good point. This was missed in an earlier Helius PR, we've been sorting by ID for a while now. You need to default to ID for people who use keyset pagination.

Another benefit: Created is the DB timestamp. Every RPC provider will have different values for created, meaning customers will have inconsistent responses between providers. ID is deterministic.

digital_asset_types/src/dao/mod.rs
NicolasPennie: Yes, a fresh deploy of DAS should not face this issue but I kept it for safety (and since we run it on our system).

nft_ingester/src/program_transformers/token/mod.rs
```diff
@@ -121,20 +121,38 @@ pub async fn handle_token_program_account<'a, 'b, 'c>(
...
+                // Update owner_type based on the supply.
+                if asset.owner_type == OwnerType::Unknown {
+                    active.owner_type = match m.supply.cmp(&1) {
+                        std::cmp::Ordering::Equal => Set(OwnerType::Single),
+                        std::cmp::Ordering::Greater => Set(OwnerType::Token),
+                        _ => NotSet,
+                    }
+                }
+
                 active.save(db).await?;
```
danenbm, line 153: Since we've recently done all these studies for race conditions due to concurrent processes, one pattern I keep an eye out for and flag as potential issue is the one where there's a separate read and write SQL statements, which is happening here.

I'm actually still ramping up on some of the token indexing, but is this scenario possible?:
1. First process: we are indexing a token with supply of 1, and one record is returned by the `find_by_id` (`OwnerType::Unknown` and `Column::Supply.eq(1)`),
2. Second (concurent) process indexing after token is burned, sets supply to 0.
3. First process would erroneously set token back to supply of 1 with the `active.save`.

Should we consider simply combining these SQL statements?

danenbm (approved):
Approved with one comment

nft_ingester/src/program_transformers/token/mod.rs
NicolasPennie: You're bang on. We should avoid doing read+write combos. One of the next tasks in our backlog is to fix the race conditions and re-work the NFT indexing, but its a much bigger and riskier project. 

The scope of this PR is to simply improve on the existing setup by adding a few retries to work around some race conditions. Also, aligning the code w/ the Helius setup will make it easier for this repo to benefit from our race condition fixes.

PR #162, opened 2024-01-23
Fix creators re-ordering bug

NicolasPennie:
## Overview 
A DB conflict would happen when the same creator would change positions in the creators array. This caused the update to drop and the NFT indexing to fail:

Original state:
```
(A, 1)
(B, 2)
(C, 3)
(D, 4)
(E, 5)
```

New data:
```
(A, 1)
(C, 2)  // This conflicts with (C, 3)
(D, 3)  // This conflicts with (D, 4)
(E, 4)  // This conflicts with (E, 5)
```

 The fix is to remove the unique constraint. We do not need it. The explanation is included in the code, under the migration changes. I'll copy them here for clarity:
 
>  We no longer want to enforce uniques on the (asset_id, creator) pairs. We may end up with duplicate (asset_id, creator) pairs during indexing, because a creator can change position. Any stale rows (older seq/slot_updated) will be ignored, meaning the API users will never see duplicate creators.

We are adding a new index on (asset_id, creator, verified) for getAssetsByCreator.

## Testing
Running in Helius prod


nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -532,70 +532,46 @@ pub async fn upsert_asset_creators<T>(
...
-    if creators.is_empty() {
-        // Bubblegum supports empty creator array.  In this case insert an empty Vec
-        // for the creator.
-        db_creators.push(asset_creators::ActiveModel {
+    let db_creators = creators
+        .iter()
+        .enumerate()
+        .map(|(i, creator)| asset_creators::ActiveModel {
             asset_id: Set(id.clone()),
-            position: Set(0),
```
danenbm, line 543: I think we still need this Bubblegum code that inserts the "empty" creator since unlike Token Metadata, Bubblegum allows an empty creator array.

nft_ingester/src/program_transformers/bubblegum/db.rs
```diff
@@ -532,70 +532,46 @@ pub async fn upsert_asset_creators<T>(
...
             ..Default::default()
-        });
-    } else {
-        // Set to prevent duplicates.
-        let mut creators_set = HashSet::new();
-
-        for (i, c) in creators.iter().enumerate() {
-            if creators_set.contains(&c.address) {
-                continue;
-            }
```
danenbm, line 558: But I think you are right and we don't need the HashSet here any more as the contracts enforce uniqueness and we don't require it.

nft_ingester/src/program_transformers/bubblegum/db.rs
NicolasPennie: Fixed. Good catch.

digital_asset_types/Cargo.toml
```diff
@@ -8,8 +8,8 @@ publish = { workspace = true }
 [dependencies]
 async-trait = { workspace = true }
 blockbuster = { workspace = true }
-borsh = { version = "0.9.3", optional = true }
-borsh-derive = { version = "0.9.3", optional = true }
+borsh = { workspace = true, optional = true  }
+borsh-derive = { workerspace = true, optional = true }
```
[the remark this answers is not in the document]
NicolasPennie: Fixed.

PR #179, opened 2024-03-27
Upstream Token Extensions From Helius

kespinola:
### Changes
- Upstream token extensions from the helius fork.
- Update program_transform after dropping plerkle as dependency of program_transforms crate.
- Update blockbuster to 2.1.0 which has parsers token extensions and mpl core.


Cargo.toml
```diff
@@ -99,6 +99,7 @@ spl-associated-token-account = ">= 1.1.3, < 3.0"
 spl-concurrent-merkle-tree = "0.2.0"
 spl-noop = "0.2.0"
 spl-token = ">= 3.5.0, < 5.0"
+spl-token-2022 = "1.0"
```
kespinola, line 102: blockbuster is also using 1.0 but 2.x is available but requires solana-sdk ~1.18

migration/src/m20240219_115532_add_extensions_column.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let connection = manager.get_connection();
+
+        connection
+            .execute(Statement::from_string(
+                DatabaseBackend::Postgres,
+                "ALTER TABLE asset ADD COLUMN mint_extensions jsonb;".to_string(),
```
danenbm, line 17: Nit: is it preferable to use the SeaORM stuff to add things, rather than raw SQL?  That's how most of the other straightforward migrations code looks when adding columns.

program_transformers/src/lib.rs
```diff
@@ -227,7 +230,18 @@ impl ProgramTransformer {
...
+                    handle_token_extensions_program_account(
+                        account_info,
+                        parsing_result,
+                        &self.storage,
+                        &self.download_metadata_notifier,
+                    )
+                    .await
+                }
+                ProgramParseResult::Bubblegum(_)
+                | ProgramParseResult::MplCore(_)
```
danenbm, line 243: Note: Needs to be updated now that MplCore is in

program_transformers/src/token/mod.rs
```diff
@@ -19,6 +19,64 @@ use {
...
+                .update_columns([
+                    token_accounts::Column::Mint,
+                    token_accounts::Column::DelegatedAmount,
+                    token_accounts::Column::Delegate,
+                    token_accounts::Column::Amount,
+                    token_accounts::Column::Frozen,
+                    token_accounts::Column::TokenProgram,
+                    token_accounts::Column::Owner,
+                    token_accounts::Column::CloseAuthority,
+                    token_accounts::Column::SlotUpdated,
```
danenbm, line 63: NIt: I like to put these in the same order as the model above.  I think it avoids things getting missed in the future since you could actually leave a column out in the OnConflict section and would not get an error.

program_transformers/src/token_extensions/mint.rs
```diff
@@ -0,0 +1,306 @@
...
+        COption::None => None,
+    };
+    let tokens_model = tokens::ActiveModel {
+        mint: Set(key_bytes.clone()),
+        token_program: Set(spl_token_program),
+        slot_updated: Set(slot),
+        supply: Set(m.account.supply as i64),
+        decimals: Set(m.account.decimals as i32),
+        close_authority: Set(None),
+        extension_data: Set(None),
```
danenbm, line 91: I know its None but its missing from the OnConflict section below.

program_transformers/src/token_extensions/mint.rs
```diff
@@ -0,0 +1,306 @@
...
+        .map_err(|e| ProgramTransformerError::SerializatonError(e.to_string()))?;
+    let asset_data_model = asset_data::ActiveModel {
+        metadata_url: Set(metadata.uri.clone()),
+        id: Set(key_bytes.clone()),
+        chain_data_mutability: Set(ChainMutability::Mutable),
+        chain_data: Set(metadata_json),
+        slot_updated: Set(slot),
+        base_info_seq: Set(Some(0)),
+        raw_name: Set(Some(metadata.name.clone().into_bytes().to_vec())),
+        raw_symbol: Set(Some(metadata.symbol.clone().into_bytes().to_vec())),
```
danenbm, line 141: Shouldn't these be included in the OnConflict section below?

program_transformers/src/token_extensions/mint.rs
```diff
@@ -0,0 +1,306 @@
...
+            nonce: Set(Some(0)),
+            seq: Set(Some(0)),
+            compressed: Set(false),
+            compressible: Set(false),
+            asset_data: Set(Some(key_bytes.clone())),
+            slot_updated: Set(Some(slot)),
+            burnt: Set(false),
+            mint_extensions: Set(Some(extensions)),
+            ..Default::default()
+        };
```
danenbm, line 241: Could this be moved down to be above the SQL execution?  Also, I don't think you will update mint_extensions in the OnConflict section. Is that intentional?

program_transformers/src/token_extensions/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use self::{mint::handle_token2022_mint_account, token_account::handle_token2022_token_account};
+
+pub async fn handle_token_extensions_program_account<'a, 'b>(
+    account_update: &AccountInfo,
+    parsing_result: &'a TokenExtensionsProgramAccount,
+    db: &'b DatabaseConnection,
+    download_metadata_notifier: &DownloadMetadataNotifier,
+) -> ProgramTransformerResult<()> {
+    match &parsing_result {
```
danenbm, line 19: Does there need to be handling for burned tokens?

kespinola: In my review now. Getting an integration test in to confirm. This is a issue with token-metadata as well.

danenbm: Token Metadata has this handling if the metadata account is empty: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/program_transformers/src/token_metadata/mod.rs#L24-L27

But then I think how DAS effectively leaves out burned accounts is because the supply is zero, like here: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/digital_asset_types/src/dao/scopes/asset.rs#L67

So then supply is probably being set by `handle_token2022_mint_account` so maybe it is already not going to return them, but they would never be marked as burnt?

kespinola: @danenbm 

> Token Metadata has this handling if the metadata account is empty: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/program_transformers/src/token_metadata/mod.rs#L24-L27

Is the account owner not set to the system program so would this account update ever come through? Looking through this EmptyAccount creation though.

> But then I think how DAS effectively leaves out burned accounts is because the supply is zero, like here: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/digital_asset_types/src/dao/scopes/asset.rs#L67

This is based on token_account same issue occurs if token account is closed and ownership transferred system program.

kespinola: I've looked into EmptyAccount which is what triggers flagging an asset as burned.

It is set within the handle_account of blockbuster when the data slice for the account is empty.

This occurs in burn_nonfungible within the token-metadata program.

https://github.com/metaplex-foundation/blockbuster/blob/9b07c07dcdbbab468c0943c5b4d83916b99912c4/blockbuster/src/programs/token_metadata/mod.rs#L73-L78

https://github.com/metaplex-foundation/mpl-token-metadata/blob/main/programs/token-metadata/program/src/utils/mod.rs#L243-L51

The token metadata account is re-assigned to the system program so does geyser send a change event under the owner of token-metadata or system program? If its the later DAS will never register the change since its account selector does not include system program.


The token extensions blockbuster parser does include the same EmptyAccount but again depends on the ownership assignment.

kespinola: On a local test validator stack when a traditional NFT is created and then the burn instruction of token-metadata is invoked. The following occurs:

1. The mint is left open but supply dropped to zero. This is correctly reflected in DAS.
2. The token account is closed and reassigned to the system program but no DAS is not notified.
3. token metadata account is left open with 1 byte. DAS receives the update event but throws 
```
{"timestamp":"2024-04-09T13:15:37.272578Z","level":"WARN","fields":{"message":"Error parsing account 5Yga8VnpA1GXnYohtH7w3NMKVczA7ZXUUA16T5UJEG16: \"Uninitialized account type\"","log.target":"nft_ingester::metrics","log.module_path":"nft_ingester::metrics","log.file":"nft_ingester/src/metrics.rs","log.line":94},"target":"nft_ingester::metrics"}
```
The asset is not marked as burnt because it the update event fails to parse.

NicolasPennie: This is because account closures go through the system program. It's annoying. We'll upstream this fix.

kespinola:
see https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/214

PR #180, opened 2024-03-27
Use separate mpl_core table

danenbm:
Normalize: using separate mpl_core table.


danenbm:
Choosing not to go this route for now see https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/178

Issue #83, opened 2023-06-21
Proposal: gRPC based indexer


linuskendall:
Currently the DAS API stack relies on having a geyser plugin indexer. One proposal would be to add/transition to a geyser grpc based indexer (https://github.com/rpcpool/yellowstone-grpc). This would allow:

1) Multiple parallel indexers connected to different grpc endpoints (even different providers)
2) Better failover
3) Lower cost since Yellowstone gRPC can be delivered as a shared service.

For account updates and transaction updates this should be pretty straight forward, you can subscribe to confirmed or finalized commitment level via gRPC, convert to Plerkle format and push to redis.

You could also implement a block check mode which checks that you have indexed all transactions from a given block and if NOT, fetch that block via gRPC to fill in missing TXs.

There is experimental support in yellowstone-grpc for fetching program account state which would allow snapshots of the Tokenkeg account to be taken for indexing of traditional NFTs. Snapshots could also be handled through one time ingest through a tool like https://github.com/terorie/solana-snapshot-etl which could even run the plerkle plugin to update all accounts at once.  With snapshot-etl the order would be:

1) Start up the geyser grpc ingestion
2) Run snapshot-etl to fill in all the missing assets while grpc subscriptions are running.

PR #183, opened 2024-04-10
GRPC Ingest

kespinola:
### Goal
Reduce the operational cost and complexity of running DAS by switching to Dragonmouth grpc stream for account and transaction updates.

Upgrade the ingestion engine by refactoring configuration, workers, and metrics.

### Approach
Create a new crate named das-grpc-ingest which has 3 primary subcommands.

stream - connects to Dragonmouth and pushes relevant events into a redis queue
ingest - subscribes to redis streams, spawns workers, and processes events using program_transformers crate.
download - fetches metadata json for asset_data replacement to bgtasks
Based on https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/144

closes https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/issues/83

## Next Steps
- Remove statsd as a dependency of program_transforms. Make the metrics backend the responsibility of the ingest runtime so prometheus can be used for grpc-ingest but nft_ingester can continue to use statsd.

## Diagram

GRPC DAS Ingest V1 drawio


.github/workflows/test.yml
```diff
@@ -25,20 +25,20 @@ jobs:
...
+            ~/.cargo/git/db/
+            digital-asset-rpc-infrastructure/target/
+          key: cargo-${{ hashFiles('**/Cargo.lock') }}-0001
 
       # Cargo.lock
       - name: Check lock file
         run: |
           cargo tree
           git checkout Cargo.lock
-          cargo tree --frozen
```
kespinola, line 41: yellowstone crates are referenced using git so require network fetch to retrieve which violates frozen flag.

niks3089:
Went over it quickly and it looks good. One question though, how would this work if we want to backfill some accounts or run account flushes to reindex some accounts?

kespinola:
> Went over it quickly and it looks good. One question though, how would this work if we want to backfill some accounts or run account flushes to reindex some accounts?

@niks3089 will continue to build out das-ops to assist in backfill tasks. For example, last week added command to flush all accounts for a program as well as reindex a single account.

I will say its has a dependency on plerkle right now but looking how to make it work for both.

https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/tree/main/ops

Cargo.toml
```diff
@@ -117,11 +130,16 @@ tokio-stream = "0.1.14"
...
 tracing-subscriber = "0.3.16"
-txn_forwarder = {path = "tools/txn_forwarder"}
+txn_forwarder = { path = "tools/txn_forwarder" }
 url = "2.3.1"
+vergen = "8.2.1"
 wasi = "0.7.0"
 wasm-bindgen = "0.2.83"
+yellowstone-grpc-client = { git = "https://github.com/rpcpool/yellowstone-grpc.git", tag = "v1.14.1+solana.1.17.28" }                          # tag is geyser plugin
+yellowstone-grpc-proto = { git = "https://github.com/rpcpool/yellowstone-grpc.git", tag = "v1.14.1+solana.1.17.28" }                           # tag is geyser plugin
+yellowstone-grpc-tools = { git = "https://github.com/rpcpool/yellowstone-grpc.git", tag = "v1.14.1+solana.1.17.28", default-features = false } # tag is geyser plugin
```
danenbm, line 142: I'm fine with this but are you holding off on publishing the yellowstone crates?

kespinola:
> im not certain it should be a wholesale switch, but a cli or config flag that allows this

Howdie @austbot. The current indexer is still available and can live side-by-side. This version ships as a separate binary (as I'm sure you see). No intention of deleting nft_ingester.

grpc-ingest/README.md
danenbm, line 1: As we discussed offline would be great if there was a more over-arching README explaining a bit of the architecture and how this is separate from the existing indexing system.  Basically you could reuse most of the content from this PR description.

danenbm (commented):
This LGTM although I haven't been able to fully deep dive all the task management details.  I'll take another look tomorrow to see if I can detect any issues, but the way everything is setup and the tasks are spawned looks clean and well-designed, so I'm thinking this is about ready to merge.  I just left a couple minor comments.

danenbm:
Thanks again for going over the desgin and architecture with me offline!

grpc-ingest/src/main.rs
```diff
@@ -0,0 +1,90 @@
...
+                .await
+                .with_context(|| format!("failed to parse config from: {}", args.config))?;
+            config.check();
+            ingester::run(config).await
+        }
+        ArgsAction::DownloadMetadata => {
+            let config = config_load::<ConfigDownloadMetadata>(&args.config)
+                .await
+                .with_context(|| format!("failed to parse config from: {}", args.config))?;
+            download_metadata::run(config).await
```
danenbm, line 87: Is there only one instance of each process running for a given data pipeline?

grpc-ingest/src/download_metadata.rs
```diff
@@ -0,0 +1,236 @@
...
+        sync::{mpsc, Notify},
+        task::JoinSet,
+        time::sleep,
+    },
+    tracing::{info, warn},
+};
+
+pub const TASK_TYPE: &str = "DownloadMetadata";
+
+pub async fn run(config: ConfigDownloadMetadata) -> anyhow::Result<()> {
```
danenbm, line 32: Related to my previous question about how many instances of each process running.  Is it safe to assume that right now there's only one instance of each type of task?  Otherwise seems like there could be race conditions such as one process is resetting tasks to `Pending` or `Running` and another process is trying to execute tasks, etc.

Another interesting option with a rearchitecture like this, you could also consider putting the download metadata tasks into Redis.

grpc-ingest/src/main.rs
kespinola: A mix. Likely want just 1 grpc.2 redis running, can have multiple ingester but better to adjust its worker count, can have multiple download metadata. 

In practice we are running 1 each with tuned thread counts.

grpc-ingest/src/download_metadata.rs
kespinola: > Another interesting option with a rearchitecture like this, you could also consider putting the download metadata tasks into Redis.

Agreed. As part of Triton running this in prod we are going to do just that based on metadata-json crate we are running now but will be combining with grpc-ingest.

https://github.com/rpcpool/digital-asset-rpc-infrastructure/tree/triton-build/metadata_json

I am also going to look at adding some metadata to asset data to help skip dead links. Idea is to add last_requested_status, requested_attempts.

nft_ingester/src/tasks/mod.rs
```diff
@@ -324,8 +324,8 @@ impl TaskManager {
         tokio::task::spawn(async move {
             while let Some(task) = receiver.recv().await {
                 if let Some(task_created_time) = task.created_at {
-                    let bus_time =
-                        Utc::now().timestamp_millis() - task_created_time.timestamp_millis();
+                    let bus_time = Utc::now().timestamp_millis()
+                        - task_created_time.and_utc().timestamp_millis();
```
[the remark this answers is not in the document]
kespinola: This was clippy recommendation.

Cargo.toml
kespinola: Let me sync with the team on this about publishing to crates.io

.github/workflows/test.yml
kespinola: https://lib.rs/crates/yellowstone-grpc-client

NVM crate is published

Cargo.toml
kespinola: https://lib.rs/crates/yellowstone-grpc-client

NVM

kespinola:
moved here

https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/240

PR #185, opened 2024-04-16
Overwrite asset authority at time of query

danenbm:
### Problem 1:
Currently when an mpl-core `Asset` is part of a collection it saves the mpl-core `Collection's` update authority as its own at the time of indexing.  This authority value becomes stale if the `Collection` update authority changes.

### Solution 1:
This PR modifies mpl-core indexing to just save the collection address on the mpl-core asset, but at the time the asset is queried, overwrite it with the current collection update authority.

An alternative solution would be to trigger a stored procedure that updates the (likely 10k) asset records if an and item with  `SpecificationAssetClass::MplCoreCollection` gets a new authority.

### Problem 2:
The solution for **Problem 1** breaks `getAssetsByAuthority` because the `Asset` does not actually have the current authority saved.  But there is another related issue with `getAssetsByCreator` for mpl-core `Assets`, because the creators might be stored on the `Collection` plugin and not even present on the `Asset`.

### Solution 2:
This PR adds additional queries in `get_by_related_condition` that looks at all the assets returned by `getAssetsByAuthority` or `getAssetsByCreator`, and if any of them have `SpecificationAssetClass::MplCoreCollection`, it does additional queries to get all the `Assets` in those collections.  It adds a couple queries and a couple SQL functions to do base58 encoding.

### Discussion
Overall looking for feedback and any other suggested solutions to the problems presented.  In my mind I think it all has to be done in one query so that the pagination works.  If we go this route I'd probably also add the additional mpl-core `Collection` query to the `searchAssets` code.

One thing this PR does not solve is overwriting collection-level info like Royalties amounts or Creators on the individual assets that are returned.  Nor does it add all the collection plugins onto the assets.  Merging all the inherited information to an mpl-core `Asset` is a bigger discussion on data representation.  There are limitations the current interface such as `frozen` status which can be updated at collection-level.

If we have more discussion on Problem 2 I could also separate it and check in the solution for Problem 1 sooner.


NicolasPennie:
Solution 1 seems the most pragmatic in the short term. Can re-evaluate if we start seeing massive Core collections or seeing collections frequently changing authority.

Solution 2 is a recipe for trouble with pagination, ordering, etc.

danenbm:
> > Solution 1 seems the most pragmatic in the short term. Can re-evaluate if we start seeing massive Core collections or seeing collections frequently changing authority.
> > Solution 2 is a recipe for trouble with pagination, ordering, etc.
> 
> I too think backfilling assets on collection update is good first approach. If we need to reindex trigger that account which then goes to update all assets associated to the colllection.
> 
> Overall prefer to front load effort on the write side to avoid complicated read queries.

I can do this, making a stored procedure that triggers the asset authority update when an MplCore collection is updated.  This will solve the staleness issue but even if that is fixed, there is still kind of a collection join issue in general.

Which is, if I call getAssetsByCreator, and the creator is only only the MplCoreCollection (which is the norm), you will only get your collection returned, and you will have to do your own additional query to get the assets in your collection.  That's where the subquery idea comes in as well.

migration/src/m20240417_120101_add_bs58_encode_function.rs
```diff
@@ -0,0 +1,108 @@
+use sea_orm_migration::{
```
kespinola, line 1: This is great btw. ty.

kespinola (commented):
In doing the backfill some items to think about:

- doing updates in batches
- also filter out who may of been updated in previous pass in case of rerun

PR #187, opened 2024-04-22
MPL Core Authority Syncing for Collections

kespinola:
resolves https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/185

### DIFF
--- a/program_transformers/src/mpl_core_program/v1_asset.rs
+++ b/program_transformers/src/mpl_core_program/v1_asset.rs
@@ -26,7 +26,8 @@ use {
         entity::{ActiveValue, ColumnTrait, EntityTrait},
         query::{JsonValue, QueryFilter, QueryTrait},
         sea_query::query::OnConflict,
-        ConnectionTrait, DbBackend, TransactionTrait,
+        sea_query::Expr,
+        ConnectionTrait, CursorTrait, DbBackend, TransactionTrait,
     },
     serde_json::{value::Value, Map},
     solana_sdk::pubkey::Pubkey,
@@ -76,8 +77,8 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
 
     // Note: This indexes both Core Assets and Core Collections.
     let asset = match account_data {
-        MplCoreAccountData::Asset(indexable_asset) => indexable_asset,
-        MplCoreAccountData::Collection(indexable_asset) => indexable_asset,
+        MplCoreAccountData::Asset(indexable_asset)
+        | MplCoreAccountData::Collection(indexable_asset) => indexable_asset,
         _ => return Err(ProgramTransformerError::NotImplemented),
     };
 
@@ -132,6 +133,11 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
     txn.execute(query)
         .await
         .map_err(|db_err| ProgramTransformerError::AssetIndexError(db_err.to_string()))?;
+
+    if matches!(account_data, MplCoreAccountData::Collection(_)) {
+        update_group_asset_authorities(conn, id_vec.clone(), update_authority.clone(), slot_i)
+            .await?;
+    }
 
     //-----------------------
     // asset_data table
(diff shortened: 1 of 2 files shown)


program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -536,3 +542,58 @@ fn convert_keys_to_snake_case(plugins_json: &mut Value) {
...
+            .col_expr(
+                asset_authority::Column::Authority,
+                Expr::value(authority.clone()),
+            )
+            .col_expr(asset_authority::Column::SlotUpdated, Expr::value(slot))
+            .filter(asset_authority::Column::AssetId.is_in(asset_ids))
+            .filter(Expr::cust_with_values(
+                "asset_authority.slot_updated <= ?",
+                vec![slot],
+            ))
```
danenbm, line 590: For some reason it doesn't like this, even though it looks like valid syntax when I ran it it never filled in the `?` with the slot value and caused a syntax error.

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -536,3 +542,58 @@ fn convert_keys_to_snake_case(plugins_json: &mut Value) {
...
+        .filter(asset_grouping::Column::GroupKey.eq(group_key.clone()))
+        .cursor_by(asset_grouping::Column::AssetId);
+    let mut query = query.first(1_000);
+
+    loop {
+        if let Some(after) = after.clone() {
+            query = query.after(after);
+        }
+
+        let entries = query.all(conn).await?;
```
danenbm, line 568: Because the query contains a `select_only`, I have found that it gets angry because `query.all()` tries to populate a whole `Model` but it only has the `AssetID` value.  

So I think we need to do something like:
```
#[derive(Debug, FromQueryResult, Clone)]
struct AssetIdQueryResult {
    pub asset_id: Vec<u8>,
}
...
let mut query = asset_grouping::Entity::find()
    .select_only()
    ...
    .into_model::<AssetIdQueryResult>();
```

which I just got from examples here: https://www.sea-ql.org/SeaORM/docs/advanced-query/custom-select/

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -536,3 +542,58 @@ fn convert_keys_to_snake_case(plugins_json: &mut Value) {
...
+    group_key: Vec<u8>,
+    authority: Vec<u8>,
+    slot: i64,
+) -> ProgramTransformerResult<()> {
+    let mut after = None;
+
+    let mut query = asset_grouping::Entity::find()
+        .select_only()
+        .column(asset_grouping::Column::AssetId)
+        .filter(asset_grouping::Column::GroupKey.eq(group_key.clone()))
```
danenbm, line 559: Group key is the word "collection" and group value is the bs58 encoded ID, so I think this needs to be:

```
let group_key = "collection".to_string();
let group_value = bs58::encode(id_vec.clone()).into_string();

let mut query = asset_grouping::Entity::find()
    .select_only()
    .column(asset_grouping::Column::AssetId)
    .filter(asset_grouping::Column::GroupKey.eq(group_key))
    .filter(asset_grouping::Column::GroupValue.eq(group_value))
    ...
```

danenbm (changes requested):
Thank you for doing this.  Now I see clearly what you meant with the cursor and batched updates!

I tested it using some typescript code that mints a few assets to a collection and then updates the collection's authority, and had a few findings, see comments.

With the changes I made according to those comments, it works.  Although for the 3rd comment I didn't actually solve it I just worked around it.  Maybe you have a better idea what's the issue with that one.

danenbm (approved):
I have tested this as well and it works great!

1. Minted 4 assets to an mpl-core collection.
2. Using `getAsset`, verified the authority was set to the collection's authority for those 4 assets, and that the collection authority was indexed.
3. Updated the collection's update authority using `updateCollectionV1` instruction on mpl-core.
4. Verified that the change was indexed by calling getAsset.  The authority was updated for the 4 assets and the collection.

One thing we may want to consider is only running the batched `update_many()` if the collection's update authority actually changed.  It would probably be slightly more complex but could save some querying.

However, right now I think it actually is more race-condition correcting in general, because asset indexing looks for the collection to get the authority (and puts empty Vec there if collection isn't found), and now collection indexing gets the assets and updates those.  Not sure if the race condition is a real issue though, as collections are created before assets are minted to them.

PR #191, opened 2024-05-02
Chore: speedy docker

stegaBOB:
Builds all the rust bins in a single builder container and copies the bins out in each of the others. Also adds docker build caches for the target directories to speed up builds


danenbm (approved):
Thank for doing this!!!

PR #192, opened 2024-05-09
Always Write CL_AUDITS_V2

kespinola:
### Background
Currently the writing of cl_audits_v2 is managed by a config option. However, getAssetSignatures requires the cl_audits_v2 table in order to return signatures history for a bubblegum asset.

As of today the cl_audits_v2 table is 170GB when there are 330GB of cl_items.

### Changes
- Always write cl_audit_v2 records by dropping the cl_audits config option.


danenbm (approved):
LGTM but might be good to hear from some others in the RPC channel as well

PR #193, opened 2024-05-15
Respect Sequential Updates of Merkle Trees

kespinola:
### Issue
In the current version of the backfiller transactions can be processed out of order from transaction responses coming out of order. The transactions are immediately pushed to redis for processing which may also result of out of order with multiple consumers processing the redis stream.

### Fix
Switch to a fan-out-fan in strategy where transaction request threads are scoped to the tree. In this updated strategy trees can be processed in parallel. Within a tree gaps and transactions threads are managed by the tree thread so max workers are now scoped. Once all transactions for the tree are retrieved sort and process in order. 

### Changes
- Resolve out of order backfill issues.
- Drop plerkle as a dependency for das-ops.
- Ops command now manage thread pools for running program transform handlers instead of pushing to a _FILL redis stream.


kespinola:
work moved here.

https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/197

PR #198, opened 2024-06-11
feat: rollup indexing

RequescoS:


program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+                inner_ix: None,
+                keys: &[],
+                slot,
+            },
+            txn,
+            "CreateTreeWithRoot",
+            cl_audits,
+        )
+        .await?;
+    }
```
danenbm, line 568: So what happens if the all these rolled mints go to `mint_v1`, but there's one missing?  Seems like that shouldn't happen because they are fully validated tree down to the merkle root.

But if it was ever possible then I think it results in a missing an entry in the `cl_audits` table, and gets detected by the backfilling process, which would try to get the missing transactions, but they wouldn't exist.  So maybe we want to prevent all rollups from being entered in the `cl_audits` table and just hardcode this `cl_audits` bool to `false`?

But then again the `cl_audits` table might also be used for auditing purposes other than gap filling, so maybe we want to put entries?  Probably needs some additional discussion.

Overall the state management is good so looks like you would have `RollupPersistingState::FailedToPersist` as an indicator that you need to retry or backfill rollups.

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+        Self {
+            txn,
+            downloader,
+            cl_audits,
+        }
+    }
+
+    pub async fn persist_rollups(&self) {
+        loop {
+            tokio::time::sleep(Duration::from_secs(5)).await;
```
danenbm, line 194: Have you considered doing a a trigger/notification when a row is added to the `rollup_to_verify` table?  That would probably work better than a hardcoded delay for the times when there are no rollups and the times there are spikes of multiple new entries.

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+                // no rollups to persist
+                continue;
+            };
+            let Ok(rollup) = rollup
+                .map(|r| bincode::deserialize::<Rollup>(r.rollup_binary_bincode.as_slice()))
+                .transpose()
+                .map_err(|e| ProgramTransformerError::DeserializationError(e.to_string()))
+            else {
+                continue;
+            };
```
danenbm, line 208: In the case of error will it keep trying to deserialize this rollup in subsequent iterations of the loop, because it is stuck in `RollupPersistingState::ReceivedTransaction` and never deleted from the table?

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+    }
+
+    pub async fn persist_rollup(
+        &self,
+        mut rollup_to_verify: rollup_to_verify::Model,
+        mut rollup: Option<Box<Rollup>>,
+    ) {
+        info!("Persisting {} rollup", &rollup_to_verify.url);
+        loop {
+            match &rollup_to_verify.rollup_persisting_state {
```
danenbm, line 221: Nice state machine!

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+                }
+                &RollupPersistingState::FailedToPersist | &RollupPersistingState::StoredUpdate => {
+                    if let Err(e) = self.drop_rollup_from_queue(&rollup_to_verify).await {
+                        error!("failed to drop rollup from queue: {}", e);
+                    };
+                    info!(
+                        "Finish processing {} rollup file with {:?} state",
+                        &rollup_to_verify.url, &rollup_to_verify.rollup_persisting_state
+                    );
+                    return;
```
danenbm, line 261: This is just a note that probably when we bring this to the wider RPC community we might want to add some specific metrics based on their ops preferences as well.

danenbm (changes requested):

Cargo.toml
```diff
@@ -93,17 +93,17 @@ sea-query = "0.28.1"
...
-solana-geyser-plugin-interface = "~1.17"
-solana-program = "~1.17"
-solana-sdk = "~1.17"
-solana-transaction-status = "~1.17"
+solana-account-decoder = "1.18.11"
+solana-client = "1.18.11"
+solana-geyser-plugin-interface = "1.18.11"
+solana-program = "1.18.11"
+solana-sdk = "1.18.11"
+solana-transaction-status = "1.18.11"
```
danenbm, line 101: Not sure if RPC community is ready to update to 1.18 FYI

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,571 @@
...
+        mut rollup_to_verify: rollup_to_verify::Model,
+        mut rollup: Option<Box<Rollup>>,
+    ) {
+        info!("Persisting {} rollup", &rollup_to_verify.url);
+        loop {
+            match &rollup_to_verify.rollup_persisting_state {
+                &RollupPersistingState::ReceivedTransaction => {
+                    if let Err(err) = self
+                        .download_rollup(&mut rollup_to_verify, &mut rollup)
+                        .await
```
danenbm, line 225: I think with most DAS processing there is an expectation that there will eventually be multiple concurrent processors of the data in order to handle volume.  So in this case, there would be multiple consumers of the same `rollup_to_verify` table row, and it seems like we would need to lock an entry or have some other means to prevent a race condition of multiple processors both downloading and then changing to the next state.  The background task processing has a lock state, which may not be perfect.  I may be misunderstanding but it seems like this processing also needs a plan for locking.

danenbm (changes requested):

blockbuster/blockbuster/src/programs/token_extensions/mod.rs
```diff
@@ -82,7 +82,6 @@ pubkeys!(
 
 pub struct Token2022AccountParser;
 
-#[allow(clippy::large_enum_variant)]
 pub enum TokenExtensionsProgramAccount {
```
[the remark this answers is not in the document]
RequescoS: Yeap, you are right, seems like the PR change was based on slightly older version of the files. Thank you

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Great catch, thank you!

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Thanks :)

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Hmm
For now i do not think it is possible to miss some asset inside rollup because of validation
And if we will find some setup that create a gap we might just add additional validation check
But for more straightforward flow sake it can be convinient to set cl_audits == false, wdyt?

Cargo.toml
RequescoS: Updated smart contracts migrated to this solana version, so we must use it here or find some detours to avoid conflicts in import versions. Or maybe we can talk with the community and notify that in the near future, it will be necessary to update the solana version?

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Thank you for noticing this! I added  RollupPersistingState::StartProcessing. So now when i select rollup to process i set state for this rollup to StartProcessing inside single transaction. i hope this will prevent such case as you described, wdyt?

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Thank you! I added listener for inserts into `rollup_to_verify` table. Primary i was thinking about share single listener between multiple instances of RollupPersister but as far as most listener methods use mutable self reference (&mut self) i decided to create new listener for each RollupPersister. i hope that together with StartProcessing state it will not cause any race conditions

RequescoS: I was also thinking about using single listener and sending notifications inside code using channels, but mpsc is very uncomfortable for such purpose because we need to send notifications for multiple consumers/workers. There also one community [crate](https://github.com/BurntSushi/chan) that implements Golang channels for Rust with multiple consumers but i think that it is bad practice to use additional import in our case, wdyt?

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Thank you. I added some metrics but not sure if i use statsd correctly (in context of nft_ingester), so i will be very appreciate if you will look at that

Cargo.toml
danenbm: ok that seems fine

program_transformers/src/rollups/rollup_persister.rs
danenbm: I think the sequence would be: if this deserialization failed, it would default to `None`, which it would pass to `persist_rollup`, which would re-download and serialize it, and update the row in the `rollup` table with new `FileHash` and `RollupBinaryBincode`, which we would assume would fix the issue.  Is this correct understanding?

But now looking at it more, when is this code even expected to run?  In the outer `persist_rollups` process, it should get a rollup to verify from the `rollup_to_verify` table and at the same time attempt to get the downloaded rollup from the `rollup` table.  But it doesn't seem like the Rollup should yet be poplulated in the rollup table until `persist_rollup` function is called which actually does the download.

program_transformers/src/rollups/rollup_persister.rs
danenbm: We can see if wider community has specific suggestions later when this PR goes to them, but as a reference implementation this looks good to me.

nft_ingester/src/main.rs
```diff
@@ -148,6 +153,18 @@ pub async fn main() -> Result<(), IngesterError> {
                 }
             }
         }
+
+        let rollup_persister = RollupPersister::new(
```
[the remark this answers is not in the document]
danenbm: Agree database config was fine.

Yes, I did mean to suggest `WorkerConfig`, so there could be a `worker_count` for number of `RollupPersister`.  Unless you really think there only needs to be one long term for the system design.  But seemed like it needed to be a scalable number.

program_transformers/src/rollups/rollup_persister.rs
danenbm: Yes I think it makes sense to set `cl_audits = false` as this will not be part of that backfill system.

program_transformers/src/rollups/rollup_persister.rs
danenbm: I think there could be a race condition within `get_rollup_to_verify()`:

1. Both processes start a transaction, and both can use a `SELECT` query to find the same row in `rollups_to_verify`, which is a row that is not in the `StartProcessing` state.

2. Both processes attempt to update the selected row in `rollups_to_verify`, which does implicitly lock the row.  The second process waits for the lock on the row to be released.

3. The first process commits the transaction and releases the lock.  The second process successfully updates the row again.

4.  Now both processes move to `persist_rollup()` and can race each other in their respective state machines.

I think to solve this issue, there's a few options.

1. You could figure out the single listener thing and then each process would receive a notification.  However, if there's multiple rows getting added at nearly the same time, then there will be multiple triggers at nearly the same time.  From what I can tell there's no specific `rollup_to_verify` row associated with a trigger, so then the multiple processes could still find the same row, causing the same concurrency issue.

2. You could use a multi-consumer/multi-producer channel.   I agree that mpsc is incorrect for this purpose.  I agree the community crate is not right either, as the README says it is end-of-life and deprecated.  The Tokio tutorial at https://tokio.rs/tokio/tutorial/channels suggests to use `async-channel` crate for multi-consumer/multi-producer, so you could look into that.

3. My recommended option is this third one, which is modify `get_rollup_to_verify()` to do a `SELECT ... FOR UPDATE`.  I believe you can accomplish this by simply adding `lock_for_update()` to the `SELECT`:
```
let rollup_to_verify = rollup_to_verify::Entity::find()
    .filter(condition)
    .order_by_asc(rollup_to_verify::Column::CreatedAtSlot)
+   .lock_for_update() // *************************** Lock the row for the current transaction
    .one(&multi_txn)
    .await
    .map_err(|e| ProgramTransformerError::DatabaseError(e.to_string()))?;
```
Then I think `persist_rollup` also needs to be modified to NOT service rollups in the `ReceivedTransaction` state.  Otherwise I think it will still grab anything that was just ingested by the program transformer.

program_transformers/src/rollups/rollup_persister.rs
danenbm: I think yes `StartProcessing` state will be able to prevent issues with multiple processors in `persist_rollup()`, but it must be slightly changed to not do any handling of items in `ReceivedTransaction` state.  See other comment for more information where I commented on `get_rollup_to_verify()` based on my study and understanding of the code.

program_transformers/src/rollups/rollup_persister.rs
RequescoS: Thank you very much!

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub last_leaf_hash: [u8; 32],
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct RolledMintInstruction {
+    pub tree_update: ChangeLogEventV1,
+    pub leaf_update: LeafSchema,
+    pub mint_args: MetadataArgs,
+    #[serde(with = "serde_with::As::<serde_with::DisplayFromStr>")]
+    pub authority: Pubkey,
```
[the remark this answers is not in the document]
RequescoS: Yeap, i think we can remove all code from sdk and just use import but in the scope of separate task. This will also add `creator_signature`

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,616 @@
...
+        mut rollup: Option<Box<Rollup>>,
+    ) {
+        let start_time = Instant::now();
+        info!("Persisting {} rollup", &rollup_to_verify.url);
+        loop {
+            match &rollup_to_verify.rollup_persisting_state {
+                &RollupPersistingState::ReceivedTransaction => {
+                    // We get ReceivedTransaction state on the start of processing
+                    rollup_to_verify.rollup_persisting_state =
+                        RollupPersistingState::StartProcessing;
```
danenbm, line 231: This case should not be possible right because anything gotten by `get_rollup_to_verify()` should be moved out of this state before being sent as a parameter to `persist_rollup`?

program_transformers/src/rollups/rollup_persister.rs
```diff
@@ -0,0 +1,616 @@
...
+                    .ne(RollupPersistingState::FailedToPersist),
+            )
+            .add(
+                rollup_to_verify::Column::RollupPersistingState
+                    .ne(RollupPersistingState::StoredUpdate),
+            )
+            .add(
+                rollup_to_verify::Column::RollupPersistingState
+                    .ne(RollupPersistingState::StartProcessing),
+            );
```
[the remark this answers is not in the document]
RequescoS: If the state of the rollup is SuccessfullyDownload or SuccessfullyValidate then it will be picked up here by processor and continue processing from this state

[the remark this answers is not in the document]
danenbm: oh yeah I see.  Somehow thought it would get reset.

danenbm (changes requested):
Unless there are significant changes, I believe I've completed my reviews and once the last open comments are resolved the change looks good to me.  I really like the thoughtful design and good testing that was added for this feature.

Next steps for this PR would be:
1. Resolve merge conflicts with main, release dependencies and update this PR to use released packages.
2. Send PR out to the wider RPC community, discuss on the RPC channel we have, get approval and then merge to main.

PR #201, opened 2024-07-19
Fix saving large token supplies

### LINKED ISSUE #200: Incorrect Supply Saved on Fungible Assets
### Issue
The column type for supply is bigint but some mints have a supply which exceeds this. There exists a `as i64` to cast the supply but instead of throwing an error this sets the number to a negative value.

https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/program_transformers/src/token/mod.rs#L113

Bonk is one such example where the supply is 9291438607932069555 outside of the range of i64.

```
spl-token display DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263

SPL Token Mint
  Address: DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263
  Program: TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
  Supply: 9291438607932069555
  Decimals: 5
  Mint authority: (not set)
  Freeze authority: (not set)
```
https://solana.fm/address/DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263

```
id                                | \xbc07c56e60ad3d3f177382eac6548fba1fd32cfd90ca02b3e7cfa185fdce7398
alt_id                            |
specification_version             | v1
specification_asset_class         | FUNGIBLE_TOKEN
owner                             |
owner_type                        | token
delegate                          |
frozen                            | f
supply                            | -9155302531447750011
supply_mint                       | \xbc07c56e60ad3d3f177382eac6548fba1fd32cfd90ca02b3e7cfa185fdce7398
compressed                        | f
compressible                      | f
```

The asset is indexed but when its time to query it this filter for only positive supply triggers.

https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/blob/main/digital_asset_types/src/dao/scopes/asset.rs#L384

### Fix

1. Switch numeric column type for supply.
2. Bring in `bigdecimal` to handle "large" numbers
3. Error in case of failed type conversion

kespinola:
### Issue
Supply is represented as a u64 on the Token but saved in the db as i64 which doesn't support the same number range. The previous code would save these large values as negative which triggers a condition in the API where supply must be > 0 . 

An example case is BONK. 
https://solana.fm/address/DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263/transactions

### Changes
- Switch DB supply columns on asset and token tables to numeric.
- Use decimal in code to represent the numeric type in postgres.

resolves #200


kespinola:
```
plerkle=> SELECT COUNT(*) FROM asset WHERE supply < 0;
 count
-------
 29508
(1 row)
```

30k assets affected.

danenbm (approved):

PR #205, opened 2024-08-15
Use bubblegum batch mint sdk for batch mint verification

n00m4d:
# What
This PR refactors code related to batch mint verification. More specificaly it uses data types and verification functions from bubblegum-batch-mint-sdk.
# Why
All the necessary data types and verification functions are already implemented on SDK side.

# IMPORTANT
Please take a closer look at changes made [here](https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/205/files#diff-567420f3fa5bd1df914a7ac618edc117510eb8f7984025f6d1a4c957880a5cd7R257) to make sure it's realy fix and not a bug, `batch_mint` changed to `r` because `batch_mint` could be only None in that place and `r` is freshly downloaded BatchMint. All the tests are passing but still looks like there was a bug.


StanChe (approved):

integration_tests/tests/integration_tests/batch_mint_tests.rs
```diff
@@ -552,7 +553,15 @@ async fn batch_mint_with_unverified_creators_test() {
     // generate batch mint with creators verified value set to true
     // but signatures will not be attached
     // batch should not be saved
-    let test_batch_mint = generate_batch_mint(10, true);
+    let mut test_batch_mint = generate_batch_mint(10);
+
+    // set creators verified to true for this test case
+    for b_mint in test_batch_mint.batch_mints.iter_mut() {
```
RequescoS, line 559: we can bring creator_verified flag setup to SDK generate_batch_mint() fn side. seems like it is not affected tree's hashes, but it may be more convenient

program_transformers/src/batch_minting/batch_mint_persister.rs
```diff
@@ -465,30 +347,49 @@ impl<T: ConnectionTrait + TransactionTrait, D: BatchMintDownloader> BatchMintPer
         batch_mint_to_verify: &mut batch_mint_to_verify::Model,
         batch_mint: &BatchMint,
     ) {
+        let collection_raw_key: Option<[u8; 32]> =
+            if let Some(key) = batch_mint_to_verify.collection.clone() {
```
RequescoS, line 351: maybe use map()?

RequescoS (approved):

integration_tests/tests/integration_tests/batch_mint_tests.rs
n00m4d: Also had such an idea but rejected it because we need verified creators only in one test but changing SDK will also require changes in other repos which use this method. So it's +2 PRs for small change, thats why decided to do it this way 🙂

program_transformers/src/batch_minting/batch_mint_persister.rs
n00m4d: the trick is that we need to return if key convertion was failed and we cannot do it from map so alternative code with map using would look like this:
```
let collection_raw_key: Option<[u8; 32]> = batch_mint_to_verify.collection.clone().map(|key| {
    match key.try_into() {
        Ok(key_as_array) => Some(key_as_array),
        Err(e) => {
            error!("Could not convert collection key received from DB: {:?}", e);

            self.mark_persisting_failed(batch_mint_to_verify).await;
            None
        }
    }
}).flatten();

if collection_raw_key.is_none() && batch_mint_to_verify.collection.is_some() {
    return;
}
```

in my opinion this refactor will not really improve code readability.

PR #214, opened 2024-11-11
Index Token Extensions and Fungible Tokens 

Nagaprasadvr:
Add support to index fungibles along with non-fungibles , so ppl can query fungible tokens as well
Add support to return token extensions data ( along with mint and token_account data)

- [x] Rebase stale branch - https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/179
- [x] fix logic
- [x] test

PR #215, opened 2024-11-14
add showFungible option flag to das

Nagaprasadvr:
- [ ] add showFungible flag to options which will be used by getAsset, getAssetsByOwner etc
- [ ] impl showFungible - match assets which are either fungibleToken or fungibleAsset

### DIFF
--- a/digital_asset_types/src/dao/scopes/asset.rs
+++ b/digital_asset_types/src/dao/scopes/asset.rs
@@ -3,10 +3,10 @@ use crate::{
         asset::{self},
         asset_authority, asset_creators, asset_data, asset_grouping, cl_audits_v2,
         extensions::{self, instruction::PascalCase},
-        sea_orm_active_enums::Instruction,
+        sea_orm_active_enums::{Instruction, SpecificationAssetClass},
         Cursor, FullAsset, GroupingSize, Pagination,
     },
-    rpc::filter::AssetSortDirection,
+    rpc::{filter::AssetSortDirection, options::Options},
 };
 use indexmap::IndexMap;
 use sea_orm::{entity::*, query::*, ConnectionTrait, DbErr, Order};
@@ -149,11 +149,21 @@ pub async fn get_assets_by_owner(
     sort_direction: Order,
     pagination: &Pagination,
     limit: u64,
-    show_unverified_collections: bool,
+    options: &Options,
 ) -> Result<Vec<FullAsset>, DbErr> {
-    let cond = Condition::all()
+    let mut cond = Condition::all()
         .add(asset::Column::Owner.eq(owner))
         .add(asset::Column::Supply.gt(0));
+
+    if options.show_fungible {
+        cond = cond.add(
+            asset::Column::SpecificationAssetClass
+                .eq(SpecificationAssetClass::FungibleToken)
+                .or(asset::Column::SpecificationAssetClass
+                    .eq(SpecificationAssetClass::FungibleAsset)),
+        );
+    }
+
     get_assets_by_condition(
         conn,
         cond,
@@ -162,7 +172,7 @@ pub async fn get_assets_by_owner(
         sort_direction,
         pagination,
         limit,
-        show_unverified_collections,
+        options.show_unverified_collections,
     )
     .await
 }
@@ -172,10 +182,21 @@ pub async fn get_assets(
     asset_ids: Vec<Vec<u8>>,
     pagination: &Pagination,
     limit: u64,
+    options: &Options,
 ) -> Result<Vec<FullAsset>, DbErr> {
-    let cond = Condition::all()
+    let mut cond = Condition::all()
         .add(asset::Column::Id.is_in(asset_ids))
         .add(asset::Column::Supply.gt(0));
+
+    if options.show_fungible {
+        cond = cond.add(
+            asset::Column::SpecificationAssetClass
+                .eq(SpecificationAssetClass::FungibleToken)
+                .or(asset::Column::SpecificationAssetClass
+                    .eq(SpecificationAssetClass::FungibleAsset)),
+        );
+    }
+
     get_assets_by_condition(
         conn,
         cond,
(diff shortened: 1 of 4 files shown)


digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -385,12 +388,19 @@ pub async fn get_by_id(
...
         asset::Entity::find_by_id(asset_id.clone()).find_also_related(asset_data::Entity);
     if !include_no_supply {
         asset_data = asset_data.filter(Condition::all().add(asset::Column::Supply.gt(0)));
     }
+
+    let token_info = if options.show_fungible {
+        get_token_by_id(conn, asset_id.clone()).await.ok()
+    } else {
+        None
+    };
```
kespinola, line 403: Can this be done by joining in this info into the asset query so there aren't 2 round trips to the db?

kespinola (changes requested):
Needs integration test.

digital_asset_types/src/dao/scopes/asset.rs
Nagaprasadvr: currently we can use find_with_related for only 2 models

digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -377,6 +379,23 @@ pub fn asset_to_rpc(asset: FullAsset, options: &Options) -> Result<RpcAsset, DbE
...
+                .mint_authority
+                .map(|s| bs58::encode(s).into_string()),
+            freeze_authority: token_info
+                .freeze_authority
+                .map(|s| bs58::encode(s).into_string()),
+            token_program: bs58::encode(token_info.token_program).into_string(),
+        })
+    } else {
+        None
+    };
```
kespinola, line 397:
```suggestion
let token_info = if options.show_fungible {
    token_info.map(|info| TokenInfo {
        supply: info.supply.try_into().unwrap_or(0),
        decimals: info.decimals as u8,
        mint_authority: info
            .mint_authority
            .map(|s| bs58::encode(s).into_string()),
        freeze_authority: info
            .freeze_authority
            .map(|s| bs58::encode(s).into_string()),
        token_program: bs58::encode(info.token_program).into_string(),
    })
} else {
    None
};
```

integration_tests/tests/integration_tests/snapshots/integration_tests__show_fungible_flag_tests__get_asset_with_show_fungible_scenario_1.snap
```diff
@@ -0,0 +1,81 @@
...
+      "verified": false
+    }
+  ],
+  "ownership": {
+    "frozen": false,
+    "delegated": false,
+    "delegate": null,
+    "ownership_model": "single",
+    "owner": ""
+  },
```
kespinola, line 70: We don't have the owner based on the fixture accounts?

kespinola (changes requested):

integration_tests/tests/integration_tests/snapshots/integration_tests__show_fungible_flag_tests__get_asset_with_show_fungible_scenario_1.snap
Nagaprasadvr: this is because we are not indexing token accounts in the fixtures

Nagaprasadvr: indexing token accounts now

integration_tests/tests/integration_tests/show_fungible_flag_tests.rs
danenbm, line 1: These tests are great but should we also include some fungible tokens like:
J1toso1uCk3RLmjorhTtrVwY9HJ7X8V9yYac6Y7kGCPn
or
something from pump.fun?

Nagaprasadvr: this would need https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/214 , so i will pull this branch and add tests scenarios for fungibles

danenbm: ah ok ty!

danenbm (approved):
LGTM

danenbm (approved):
This LGTM.  After I approve everything I'm also going to ask Helius to take a look at all these before we merge them into main.  It might be good to just combine 214 and 215 in some way so that same content needs to be reviewed twice by them.

Or maybe its just as simple as I will let them know 214 already contained in 215.

digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -348,36 +349,32 @@ pub fn asset_to_rpc(asset: FullAsset, options: &Options) -> Result<RpcAsset, DbE
...
+    let uses = data.chain_data.get("uses").map(|u| Uses {
+        use_method: u
+            .get("use_method")
+            .and_then(|s| s.as_str())
+            .unwrap_or("Single")
+            .to_string()
+            .into(),
+        total: u.get("total").and_then(|t| t.as_u64()).unwrap_or(0),
+        remaining: u.get("remaining").and_then(|t| t.as_u64()).unwrap_or(0),
+    });
```
kespinola, line 377: In #214 data was changed to options because fungible assets may not have associated metadata but after reviewing variations all on-chain token standards TE, core, TokenK + Metdata provide methods for populating the `asset_data` thus DAS will now return assets that are "fungible" but only if they have metadata information.

kespinola (approved):

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -385,19 +396,25 @@ pub async fn get_by_id(
...
+    let (asset, data): (asset::Model, asset_data::Model) =
         asset_data.one(conn).await.and_then(|o| match o {
-            Some((a, d)) => Ok((a, d)),
+            Some((a, Some(d))) => Ok((a, d)),
```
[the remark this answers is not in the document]
kespinola: related to #231

digital_asset_types/src/dapi/common/asset.rs
danenbm: That seems fine with me.  Some don't have token standards is the main thing I think of with old fungible assets

digital_asset_types/src/rpc/asset.rs
```diff
@@ -391,6 +391,17 @@ pub struct TokenInscriptionInfo {
...
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct TokenInfo {
+    pub supply: u64,
+    pub decimals: u8,
+    pub token_program: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub mint_authority: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub freeze_authority: Option<String>,
+}
```
danenbm, line 403: I think Helius is also adding in token price in USDC here.  I know that's like some kind of extra Oracle lookup at query time, but just wanted to call it out as an FYI.

```
"token_info" : {
         "decimals" : 9,
         "mint_authority" : "6iQKfEyhr3bZMotVkW6beNZz5CPAkiwvgV2CTje9pVSS",
         "price_info" : {
            "currency" : "USDC",
            "price_per_token" : 247.02798
         },
         "supply" : 12572360539785052,
         "symbol" : "JitoSOL",
         "token_program" : "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
}
```

danenbm (approved):

digital_asset_types/src/rpc/asset.rs
kespinola: We noted this but if its requires third party API call I wanted to leave out. I'd personally like everything be from the chain.

kespinola (approved):

PR #216, opened 2024-11-15
Add getNftEditions rpc method to das

Nagaprasadvr:
- [x] add getNftEditions rpc method
- [x] pass a master edition mint to fetch editions accounts associated with master edition and return them
- [x] add integration test


program_transformers/src/token_metadata/master_edition.rs
```diff
@@ -104,5 +121,6 @@ pub async fn save_master_edition(
         )
         .build(DbBackend::Postgres);
```
kespinola, line 122: Need to add the only update when slot updated larger clause.

program_transformers/src/token_metadata/v1_asset.rs
```diff
@@ -401,6 +412,11 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
     }
     txn.commit().await?;
 
+    // If the asset is a non-fungible token, then we need to insert to the asset_v1_account_attachments table
+    if check_is_non_fungible_from_token_standard(metadata.token_standard) {
+        upsert_asset_v1_account_attachments(conn, &mint_pubkey, slot).await?;
+    }
```
kespinola, line 418:
```suggestion
    if Some(true) = metadata.token_standard.map(|token_standard| token_standard.is_non_fungible()) {
        upsert_asset_v1_account_attachments(conn, &mint_pubkey, slot).await?;
    }
```

kespinola: Need to implement is_non_fungible() but I think this composes better.

program_transformers/src/token_metadata/v1_asset.rs
```diff
@@ -411,3 +427,34 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+    let query = asset_v1_account_attachments::Entity::insert(attachment)
+        .on_conflict(
+            OnConflict::columns([asset_v1_account_attachments::Column::Id])
+                .update_columns([
+                    asset_v1_account_attachments::Column::SlotUpdated,
+                    asset_v1_account_attachments::Column::AssetId,
+                ])
+                .to_owned(),
+        )
+        .build(DbBackend::Postgres);
```
kespinola, line 455: Since the mint is never updateable we should only only be adding the asset_id. Leave the slot updated to the edition.

kespinola (changes requested):

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -553,3 +561,79 @@ fn filter_out_stale_creators(creators: &mut Vec<asset_creators::Model>) {
...
+                .eq(V1AccountAttachments::Edition)
+                .and(asset_v1_account_attachments::Column::Data.is_not_null())
+                .and(Expr::cust(&format!(
+                    "data->>'parent' = '{}'",
+                    master_edition_pubkey
+                ))),
+        )
+        .order_by_asc(asset_v1_account_attachments::Column::SlotUpdated)
+        .all(conn)
+        .await?;
```
kespinola, line 607: The limit and offset should be used to page through the editions. There is some helpers in the api crate for handling pagination.

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -553,3 +561,79 @@ fn filter_out_stale_creators(creators: &mut Vec<asset_creators::Model>) {
...
+
+    let nft_editions = nft_editions
+        .iter()
+        .map(|e| -> Result<NftEdition, DbErr> {
+            let data: Edition = e
+                .data
+                .clone()
+                .ok_or(DbErr::RecordNotFound("Edition data not found".to_string()))
+                .map(get_edition_data_from_json)??;
```
kespinola, line 616: Its normal to chain the map to the query result and to implement a TryInto for the conversion so code looks like

```
    let nft_editions = asset_v1_account_attachments::Entity::find()
        .filter(
            asset_v1_account_attachments::Column::AttachmentType
                .eq(V1AccountAttachments::Edition)
                .and(asset_v1_account_attachments::Column::Data.is_not_null())
                .and(Expr::cust(&format!(
                    "data->>'parent' = '{}'",
                    master_edition_pubkey
                ))),
        )
        .order_by_asc(asset_v1_account_attachments::Column::SlotUpdated)
        .all(conn)
        .await?
        .into_iter()
        .map(TryInto::try_into)
        .collect::<Result<..>>()?;

```

kespinola (changes requested):
Primarily need to account for pagination and could you please leave some notes on the 2-step indexing to link master edition to an edition for the readers?

program_transformers/src/token_metadata/mod.rs
```diff
@@ -45,9 +49,32 @@ pub async fn handle_token_metadata_account<'a, 'b>(
...
         // TokenMetadataAccountData::CollectionAuthorityRecord(_) => {}
         _ => Err(ProgramTransformerError::NotImplemented),
     }
 }
+
+pub trait IsNonFungibeFromTokenStandard {
+    fn is_non_fungible(&self) -> bool;
+}
+
+impl IsNonFungibeFromTokenStandard for TokenStandard {
```
danenbm, line 70: Nit: spelling of trait name

program_transformers/src/token_metadata/v1_asset.rs
```diff
@@ -411,3 +418,31 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+    slot: u64,
+) -> ProgramTransformerResult<()> {
+    let edition_pubkey = MasterEdition::find_pda(mint_pubkey).0;
+    let mint_pubkey_vec = mint_pubkey.to_bytes().to_vec();
+    let attachment = asset_v1_account_attachments::ActiveModel {
+        id: ActiveValue::Set(edition_pubkey.to_bytes().to_vec()),
+        asset_id: ActiveValue::Set(Some(mint_pubkey_vec.clone())),
+        slot_updated: ActiveValue::Set(slot as i64),
+        // by default, the attachment type is MasterEditionV1
+        attachment_type: ActiveValue::Set(V1AccountAttachments::MasterEditionV1),
```
danenbm, line 434: You could probably assume `MasterEditionV2` instead as way more of them are V2 now.  Although I know its planned to be updated by the other program transformer handler anyways.

Unless did you have some specific reason for choosing V1 as a default?  V2 is a subset of V1, but doesn't seem like that would matter here.

program_transformers/src/token_metadata/master_edition.rs
```diff
@@ -85,14 +75,47 @@ pub async fn save_master_edition(
         ..Default::default()
     };
 
-    if let Some((_me, Some(asset))) = master_edition {
-        let mut updatable: asset::ActiveModel = asset.into();
-        updatable.supply = ActiveValue::Set(Decimal::from(1));
-        updatable.specification_asset_class = ActiveValue::Set(Some(SpecificationAssetClass::Nft));
```
danenbm, line 91: This is strange I don't understand why this used to be here.  Seems really extra.

danenbm (approved):
Overall looks great!!!  Approved with just a couple questions/nit comment.

program_transformers/src/token_metadata/v1_asset.rs
Nagaprasadvr: no specific reason , will set it to MasterEditionV2 as default

program_transformers/src/token_metadata/master_edition.rs
Nagaprasadvr: yeah, this should be done when metadata acc is created or updated

program_transformers/src/token_metadata/mod.rs
Nagaprasadvr: changed to IsNonFungible

kespinola (approved):

PR #218, opened 2024-11-15
Added Token Type Filter in search_assets fuction

AhzamAkhtar:
Types of filters that are added in search_assets function are
- [x] Fungible
- [x] NonFungible
- [x] Compressed
- [x] Nft
- [x] All
- [x] Tests


digital_asset_types/src/dao/mod.rs
```diff
@@ -93,6 +96,23 @@ impl SearchAssetsQuery {
                     .clone()
                     .map(|x| asset::Column::SpecificationAssetClass.eq(x)),
             )
+            .add_option(
```
Nagaprasadvr, line 99: i guess we can remove specification_asset_class option now as we have token_type

digital_asset_types/src/dao/mod.rs
```diff
@@ -54,6 +56,7 @@ pub struct SearchAssetsQuery {
     pub condition_type: Option<ConditionType>,
     pub specification_version: Option<SpecificationVersions>,
     pub specification_asset_class: Option<SpecificationAssetClass>,
+    pub token_type: Option<TokenTypeClass>,
```
Nagaprasadvr, line 59: remove SpecificationAssetClass option

Nagaprasadvr (commented):
looks good , we can remove specification_asset_class in the query params

digital_asset_types/src/dao/mod.rs
AhzamAkhtar: yeah think so.

kespinola (changes requested):
Looks good but needs integration tests @AhzamAkhtar

AhzamAkhtar:
> Looks good but needs integration tests @AhzamAkhtar

Integration tests added @kespinola

kespinola (approved):

PR #220, opened 2024-11-18
Added option to display showNativeBalance

AhzamAkhtar:
Functions that are supported to show showNativeBalance option

1.searchAssets
2.getAssetByOwner


kespinola (changes requested):
Checking to see if we can get this in the normal index flow. DAS API should not be dependent on solana rpc.

Also a note to add integration tests.

PR #221, opened 2024-11-19
Index token inscriptions to asset_v1_account_attachments table

Nagaprasadvr:
- [x] impl indexing token_inscriptions from libreplex lib
- [x] save indexed data to asset_v1_account_attachments table inside data field and map asset_id to mint
- [ ] Add tests


Nagaprasadvr:
closing this,will open a new with filters via options

PR #227, opened 2024-11-29
Remove duplicate results for same asset_ids in get_assets rpc method

AhzamAkhtar:
**Problem**
In getAssets rpc method , a user can provide multiple same asset_id to search for, and the api will give that many duplicated results, instead we can return a single result.

Tasks
- [x] Remove duplicate results
- [x] tests


das_api/src/api/api_impl.rs
```diff
@@ -230,9 +241,13 @@ impl ApiContract for DasApi {
 
         let options = options.unwrap_or_default();
 
-        let assets = get_assets(&self.db_connection, id_bytes, batch_size as u64, &options).await?;
+        let unique_asset_ids =  remove_duplicates_asset_ids(id_bytes);
```
Nagaprasadvr, line 244: filter dup ids at line no 237 and then convert to Vec<Vec<_>>

das_api/src/api/api_impl.rs
```diff
@@ -148,6 +149,16 @@ pub fn not_found(asset_id: &String) -> DbErr {
     DbErr::RecordNotFound(format!("Asset Proof for {} Not Found", asset_id))
 }
 
+pub fn remove_duplicates_asset_ids(id_bytes: Vec<Vec<u8>>) -> Vec<Vec<u8>> {
```
Nagaprasadvr, line 152: can use generics here

Nagaprasadvr (changes requested):

das_api/src/api/api_impl.rs
```diff
@@ -218,6 +224,7 @@ impl ApiContract for DasApi {
     ) -> Result<Vec<Option<Asset>>, DasApiError> {
         let GetAssets { ids, options } = payload;
 
+        let ids = remove_duplicates_ids(ids);
```
AhzamAkhtar, line 227: also dont need to use generic now cause now we filtering only ids, which is Vec<String>

danenbm (approved):
LGTM

Nagaprasadvr (approved):

PR #229, opened 2024-12-09
Avoid unnecessary operations if Vec<Asset> is empty

AhzamAkhtar:
- [x] Avoid unnecessary operation calls if **assets** vector is empty.

### DIFF
--- a/digital_asset_types/src/dapi/common/asset.rs
+++ b/digital_asset_types/src/dapi/common/asset.rs
@@ -74,20 +74,20 @@ pub fn build_asset_response(
     let base_asset_list = AssetList {
         total,
         limit: limit as u32,
-        page : page.map(|x| x as u32),
+        page: page.map(|x| x as u32),
         before,
         after,
         items: vec![],
         errors: vec![],
         cursor,
     };
-    
+
     if assets.is_empty() {
         return base_asset_list;
     }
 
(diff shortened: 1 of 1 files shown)


digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -71,16 +71,27 @@ pub fn build_asset_response(
         }
     };
 
-    let (items, errors) = asset_list_to_rpc(assets, options);
-    AssetList {
+    let base_asset_list = AssetList {
```
Nagaprasadvr, line 74: can we move this check inside asset_list_to_rpc ?

Nagaprasadvr (changes requested):

digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -71,16 +71,27 @@ pub fn build_asset_response(
...
         total,
         limit: limit as u32,
-        page: page.map(|x| x as u32),
+        page : page.map(|x| x as u32),
```
danenbm, line 77: NIt: whitespace

danenbm (approved):

danenbm (approved):

kespinola:
There are no async calls as a result of rpc transform so I'm going to close this issue. If introduced in the future should have a benchmark to note time/utilization improvement.

PR #230, opened 2024-12-10
WIP : optimize search assets query

Nagaprasadvr:
- [ ] Add composite indexes on asset table
- [ ] Add composite indexes on asset_grouping table
- [ ] drop order by for non-numeric fields which increased latency


Nagaprasadvr:
linking this in https://github.com/rpcpool/digital-asset-rpc-infrastructure

PR #231, opened 2025-01-07
add download json_data task for token metdata extensions

Nagaprasadvr:
- [x] Run download_metdata_json background task to download json data from uri


program_transformers/src/token_extensions/mod.rs
```diff
@@ -257,7 +262,10 @@ async fn upsert_asset_data(
 
     txn.commit().await?;
 
-    Ok(())
+    Ok(DownloadMetadataInfo {
+        uri: metadata.uri.clone(),
+        asset_data_id: key_bytes,
+    })
```
kespinola, line 268: I don't see the guarantee metadata uri will be defined. Or it may be empty which isn't good either. Should DownloadMetadataInfo be returned if the metadata extension is used?

program_transformers/src/token_extensions/mod.rs
```diff
@@ -195,7 +196,11 @@ pub async fn handle_token_extensions_program_account<'a, 'b, 'c>(
             txn.commit().await?;
 
             if let Some(metadata) = &extensions.metadata {
-                upsert_asset_data(metadata, account_key.clone(), slot, db).await?;
+                if let Ok(info) = upsert_asset_data(metadata, account_key.clone(), slot, db).await {
+                    download_metadata_notifier(info)
+                        .await
+                        .map_err(ProgramTransformerError::DownloadMetadataNotify)?;
+                }
```
kespinola, line 203:
```suggestion
                upsert_asset_data(metadata, account_key.clone(), slot, db)
                  .then(download_metadata_notifier)
                  .await
                  .map_err(ProgramTransformerError::DownloadMetadataNotify)?;
                }
```
I think you can chain those calls.

kespinola (changes requested):
Looking to confirm metadata uri is always set and shouldn't be no-op unless metadata plugin is present.

program_transformers/src/token_extensions/mod.rs
Nagaprasadvr: added url is empty checks

program_transformers/src/token_extensions/mod.rs
Nagaprasadvr: as we have an Option now it can be dificult to read wyt ?

danenbm (approved):
LGTM

PR #232, opened 2025-01-08
change extension bytes data to base58 encoded

Nagaprasadvr:
- [x] token extensions has confidential transfers and it involes elgamal pubkey, cipher texts which are structured in bytes format , convert these to readable base58 format 

**`Note` : changed docker db postgres volume setup from `./db-data/:/var/lib/postgresql/data/:rw` to `./db-data/:/var/lib/postgresql/data/:z `which adds support for SELinux-enabled systems like fedora**


kespinola (approved):

docker-compose.yaml
```diff
@@ -87,16 +87,25 @@ services:
...
+      ]
     ports:
       - 5432:5432
     environment:
       POSTGRES_USER: solana # The PostgreSQL user (useful to connect to the database)
       POSTGRES_PASSWORD: solana # The PostgreSQL password (useful to connect to the database)
       POSTGRES_DB: solana
     volumes:
-      - ./db-data/:/var/lib/postgresql/data/:rw
+      - ./db-data/:/var/lib/postgresql/data/:z
```
danenbm, line 108: Why do we switch it to 'z'?

blockbuster/src/programs/token_extensions/extension.rs
```diff
@@ -216,114 +187,6 @@ pub struct ShadowMetadata {
...
-    {
-        deserializer.deserialize_bytes(ShadowElGamalCiphertextVisitor)
-    }
-}
-
-impl Default for ShadowElGamalCiphertext {
-    fn default() -> Self {
-        ShadowElGamalCiphertext([0u8; ELGAMAL_CIPHERTEXT_LEN])
-    }
-}
```
danenbm, line 262: Assuming all this was just not needed because the extensions are actually deserialized elsewhere using functions such as `get_extension()` etc.

danenbm (approved):
LGTM with couple minor questions

blockbuster/src/programs/token_extensions/extension.rs
Nagaprasadvr: Yeah , few extras here

docker-compose.yaml
Nagaprasadvr: I wil roll it back , I was in fedora and rw was not working for me

danenbm: I'm ok with it was just curious why needed

Nagaprasadvr: SELinux (security enhanced) based linux systems needs this z mode to work like fedora latest one

PR #233, opened 2025-01-08
token_type filter added in search_asset rpc method

AhzamAkhtar:
1. Types of filters that are added in search_assets function are
- [x] Fungible
- [x] NonFungible
- [x] Compressed
- [x] Nft
- [x] All
2. Search_asset rpc call skipping fungible_assets bug fixed
3. Not able to search for fungible asset using specific `owner_address` in search_asset bug fixed
4. Added validation errors and fixed search_assets compatibility issues
- [x] `owner_type` and `token_type` parameter can't be use together
- [x]  In search_assets, SearchAssetQuery doesn't support `interface` fixed.
- [x] `token_type` filter should only be used using `owner_address`
- [x] `token_type` filter can't be used with `interface`
5. Tests


digital_asset_types/src/dao/mod.rs
```diff
@@ -88,16 +114,36 @@ impl SearchAssetsQuery {
                     .clone()
                     .map(|x| asset::Column::SpecificationVersion.eq(x)),
             )
+            .add_option({
+                self.validation_checks()?;
```
kespinola, line 118: We should verify before this point and add_options always or use this check for deciding to include or not. Not for correctness of params.

digital_asset_types/src/dao/mod.rs
```diff
@@ -88,16 +116,35 @@ impl SearchAssetsQuery {
...
+                                .eq(SpecificationAssetClass::MplCoreAsset))
+                            .or(asset::Column::SpecificationAssetClass
+                                .eq(SpecificationAssetClass::ProgrammableNft)),
+                    ),
+                    TokenTypeClass::NonFungible => asset::Column::SpecificationAssetClass
+                        .eq(SpecificationAssetClass::Nft)
+                        .or(asset::Column::SpecificationAssetClass
+                            .eq(SpecificationAssetClass::ProgrammableNft))
+                        .or(asset::Column::SpecificationAssetClass
+                            .eq(SpecificationAssetClass::MplCoreAsset)),
```
danenbm, line 135: I think for both of these match arms `SpecificationAssetClass::MplCoreCollection` is missing.

integration_tests/tests/integration_tests/snapshots/integration_tests__regular_nft_tests__reg_get_asset_batch-only-2.snap
```diff
@@ -1,18 +1,17 @@
...
-    "id": "HTKAVZZrDdyecCxzm3WEkCsG1GUmiqKm73PvngfuYRNK",
+    "id": "2NqdYX6kJmMUoChnDXU2UrP9BsoPZivRw3uJG8iDhRRd",
     "content": {
       "$schema": "https://schema.metaplex.com/nft1.0.json",
-      "json_uri": "https://madlads.s3.us-west-2.amazonaws.com/json/9965.json",
+      "json_uri": "https://madlads.s3.us-west-2.amazonaws.com/json/9959.json",
       "files": [],
       "metadata": {
-        "name": "Mad Lads #9965",
+        "name": "Mad Lads #9959",
```
danenbm, line 14: Why did this test result change order?  The order for "only-2" didn't change in regular_nft_tests.rs, and I didn't see any new conditions added for `get_assets`.

danenbm (changes requested):

integration_tests/tests/integration_tests/snapshots/integration_tests__regular_nft_tests__reg_get_asset_batch-only-2.snap
Nagaprasadvr: there was a sorting logic which wasnt needed removed it now

kespinola (approved):

danenbm (approved):
approved LGTM!

PR #234, opened 2025-01-23
Display supply object for more interfaces

n00m4d:
# What

This PR expands the range of interfaces where the supply object is included in the API response. Essentially, additional interfaces were added that may contain edition information, which can now be incorporated into the supply object.

The changes in the integration test snapshots provide a clear view of how these updates impact the user experience.

# Important

This PR also changes one of the snapshots in integration tests which related to old changes merged in main already. It's `integration_tests/tests/integration_tests/snapshots/integration_tests__show_fungible_flag_tests__get_asset_with_show_fungible_scenario_3.snap` file. Looks like there were made some changes but integration tests were not updated.


danenbm (approved):
LGTM, added others to review for visibility

digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -503,7 +503,11 @@ pub fn asset_to_rpc(asset: FullAsset, options: &Options) -> Result<RpcAsset, DbE
...
         }),
         supply: match interface {
-            Interface::V1NFT => Some(Supply {
+            Interface::V1NFT
+            | Interface::LEGACY_NFT
+            | Interface::Nft
+            | Interface::ProgrammableNFT
+            | Interface::Custom => Some(Supply {
                 edition_nonce,
                 print_current_supply: 0,
```
Nagaprasadvr, line 512: i guess we should not be hardcoding this to 0 and rather fetch them from master_edition and asset_data

Nagaprasadvr (approved):
looks good , added a comment to fix in future

digital_asset_types/src/dapi/common/asset.rs
n00m4d: thanks! ticket is created, will do in a future

PR #235, opened 2025-01-29
Optimize searchAssets and getAssetsByOwner and other queries 

Nagaprasadvr:
- [ ] remove default order_by using id ( very expensive )
- [ ] remove sort_by id option for all queries


kespinola:
@Nagaprasadvr I reviewed this some and our default sort by is ID (where ID is a pubic) or pubkey. This isn't going to preserve ordering we need to find a column default and make sure its indexed. All the sort columns need indexes as well.


```
    let token_accounts = paginate(
        pagination,
        limit,
        token_accounts::Entity::find().filter(condition),
        Order::Asc,
        token_accounts::Column::Pubkey,
    )
    .all(conn)
    .await?;
```

Flagging this @danenbm if you have opinion on what column we should use for default pagination sorts. slot DESC seems like a good candidate or we can used the internal created_at so we return oldest results to the user and start working toward newer records.

danenbm:
> @Nagaprasadvr I reviewed this some and our default sort by is ID (where ID is a pubic) or pubkey. This isn't going to preserve ordering we need to find a column default and make sure its indexed. All the sort columns need indexes as well.
> 
> ```rust
>     let token_accounts = paginate(
>         pagination,
>         limit,
>         token_accounts::Entity::find().filter(condition),
>         Order::Asc,
>         token_accounts::Column::Pubkey,
>     )
>     .all(conn)
>     .await?;
> ```
> 
> Flagging this @danenbm if you have opinion on what column we should use for default pagination sorts. slot DESC seems like a good candidate or we can used the internal created_at so we return oldest results to the user and start working toward newer records.

Let's also get input from @StanChe

kespinola:
> Wonder does that change whether we deprecate sorting by ID

@danenbm switch to cursor doesn't remove the need to sort by consistent column. 

We also want to make sure we only sort when its matters that is when we paginate. No need to sort on GET queries.

kespinola:
@danenbm FYI that this doesn't remove the ID as the default pagination options but we would like to follow up on this and get PR up that switches ensures all methods require a sensible sort column.

kespinola (approved):

StanChe:
I see a potential issue with removing the sorting by id. The requests and pagination will become non-deterministic if we sort only by non-unique fields like slot_created or slot_updated. Given that multiple assets can be updated in a single slot, two identical requests might return different results. This affects both page-based pagination (which I'd discourage using) and cursor-based pagination.
While sorting by id might not be very user-centric (as most use cases I know of rely on created/updated dates), it provides crucial determinism. 
What specific requests are causing performance issues with the id sorting? Understanding this would help find an optimization that doesn't sacrifice consistency.

kespinola:
> I see a potential issue with removing the sorting by id. The requests and pagination will become non-deterministic if we sort only by non-unique fields like slot_created or slot_updated. Given that multiple assets can be updated in a single slot, two identical requests might return different results. This affects both page-based pagination (which I'd discourage using) and cursor-based pagination. While sorting by id might not be very user-centric (as most use cases I know of rely on created/updated dates), it provides crucial determinism. What specific requests are causing performance issues with the id sorting? Understanding this would help find an optimization that doesn't sacrifice consistency.

@StanChe good point on determinism. Sorting purely by slot (or any non-unique field) can lead to pagination inconsistencies if multiple assets share the same value.

One approach is to use cursor-based pagination with a deterministic sort order:
- Sort by slot DESC, id ASC to ensure stable results.
- Index the sort order with CREATE INDEX ON token_accounts (slot DESC, id ASC); to optimize queries.
- This balances performance and determinism while returning results in chronological order.

We should avoid offset-based pagination where possible, as it doesn’t scale well with large datasets so happy to open series of PRs to remove page based pagination.

danenbm:
> > I see a potential issue with removing the sorting by id. The requests and pagination will become non-deterministic if we sort only by non-unique fields like slot_created or slot_updated. Given that multiple assets can be updated in a single slot, two identical requests might return different results. This affects both page-based pagination (which I'd discourage using) and cursor-based pagination. While sorting by id might not be very user-centric (as most use cases I know of rely on created/updated dates), it provides crucial determinism. What specific requests are causing performance issues with the id sorting? Understanding this would help find an optimization that doesn't sacrifice consistency.
> 
> @StanChe good point on determinism. Sorting purely by slot (or any non-unique field) can lead to pagination inconsistencies if multiple assets share the same value.
> 
> One approach is to use cursor-based pagination with a deterministic sort order:
> 
> * Sort by slot DESC, id ASC to ensure stable results.
> * Index the sort order with CREATE INDEX ON token_accounts (slot DESC, id ASC); to optimize queries.
> * This balances performance and determinism while returning results in chronological order.
> 
> We should avoid offset-based pagination where possible, as it doesn’t scale well with large datasets so happy to open series of PRs to remove page based pagination.

People (especially users that are new to DAS and don't have many items to query) like to start with the page-based.  We did some testing and found that a page limit of 50 was sufficient to not hurt performance too much.  We could even reduce that further, but I think removing it outright could generate some frustration from users

danenbm:
> > Wonder does that change whether we deprecate sorting by ID
> 
> @danenbm switch to cursor doesn't remove the need to sort by consistent column.
> 
> We also want to make sure we only sort when its matters that is when we paginate. No need to sort on GET queries.

I guess I was thinking if ordering by ID causes performance issues, limiting pagination to a small number of pages might mitigate it enough to where its not a problem. But maybe I'm conflating topics

danenbm:
> > I see a potential issue with removing the sorting by id. The requests and pagination will become non-deterministic if we sort only by non-unique fields like slot_created or slot_updated. Given that multiple assets can be updated in a single slot, two identical requests might return different results. This affects both page-based pagination (which I'd discourage using) and cursor-based pagination. While sorting by id might not be very user-centric (as most use cases I know of rely on created/updated dates), it provides crucial determinism. What specific requests are causing performance issues with the id sorting? Understanding this would help find an optimization that doesn't sacrifice consistency.
> 
> @StanChe good point on determinism. Sorting purely by slot (or any non-unique field) can lead to pagination inconsistencies if multiple assets share the same value.
> 
> One approach is to use cursor-based pagination with a deterministic sort order:
> 
> * Sort by slot DESC, id ASC to ensure stable results.
> * Index the sort order with CREATE INDEX ON token_accounts (slot DESC, id ASC); to optimize queries.
> * This balances performance and determinism while returning results in chronological order.
> 
> We should avoid offset-based pagination where possible, as it doesn’t scale well with large datasets so happy to open series of PRs to remove page based pagination.

We do have users that use page-based pagination.  I think users who don't have a lot of assets to display and are new to DAS often start with it as its intuitive.

@StanChe did some testing and determined that a page limit of 50 was a reasonable compromise between usability and performance, at least on Aura.  We could reduce that even further probably.

I'm not fully opposed to deprecate page-based paginations, just slightly concerned that removing it may cause frustrations,  and if Helius, Triton, and Aura don't all remove it at the same time, people will get confused on why its a feature for some RPCs and not others.

kespinola:
We have some changes to the base queries that we'll open but have available in our fork now.

PR #240, opened 2025-02-10
cleanup grpc-ingest branch and limit external non-dep code changes

Nagaprasadvr:


Cargo.toml
```diff
@@ -94,23 +107,25 @@ sea-orm-migration = "0.10.6"
...
+serde_yaml = "0.9.34"
 serial_test = "2.0.0"
-solana-account-decoder = "~1.18.15"
-solana-client = "~1.18.15"
-solana-geyser-plugin-interface = "~1.18.15"
-solana-program = "~1.18.15"
-solana-sdk = "~1.18.15"
-solana-transaction-status = "~1.18.15"
-solana-zk-token-sdk = "~1.18.15"
-spl-account-compression = "0.4.2"
```
kespinola, line 105: This seems fine to keep what is in main.

README.md
```diff
@@ -15,6 +15,17 @@ super-fast querying and searching, as well as serves the merkle proofs needed to
 The API specification is located here https://github.com/metaplex-foundation/api-specifications
 This spec is what providers of this api must implement against.
 
+#### GRPC-INGEST [/grpc-ingest/README.md](/grpc-ingest/README.md)
+
+This repo also houses the GRPC-INGEST component. This is a reimplementation of nft-ingester
+This components separates grpc and ingester into two separate components. This is to allow for more flexibility in the future also increases the performance of the system.
```
kespinola, line 21:
```suggestion
The grc ingester allows the DAS index stack to receive solana change events from a Yellowstone GRPC endpoint. All the largest RPC providers offer this API so you are able to use a shared event stream instead of running a dedicated validator like you have to with the plerkle geyser plugin.
```

README.md
```diff
@@ -15,6 +15,17 @@ super-fast querying and searching, as well as serves the merkle proofs needed to
...
+
+The two components are:
+
+- grpc2redis - This component listens to a gRPC stream (triton's Dragon's Mouth gRPC) for account and transaction updates and pushes them to a pipeline that flushes data to redis at regular intervals in real time.
+
+- ingester - This component listens to the redis stream and processes the data using program transformers. It also downloads token metadata json and stores them in a postgres db using sea-orm.
```
kespinola, line 27: Please remove from the main read me but make sure the contents are covered in the grpc-ingeest README.

grpc-ingest/README.md
```diff
@@ -0,0 +1,54 @@
...
+RUST_LOG=info cargo run --bin das-grpc-ingest  -- --config grpc-ingest/config-grpc2redis.yml grpc2redis
+```
+
+### Config for Ingester [./config-ingester.yml](./config-ingester.yml)
+
+### Run the Ingester service
+
+This service performs many concurrent tasks
+
+- Fetch account updates from redis and process them using using program_transformer
```
kespinola, line 43:
```suggestion
- Fetch account updates from redis and process them using the `program_transformer` crate
```

program_transformers/src/bubblegum/db.rs
```diff
@@ -435,9 +434,7 @@ where
                     asset_data::Column::ChainData,
                     asset_data::Column::MetadataUrl,
                     asset_data::Column::MetadataMutability,
-                    // Don't update asset_data::Column::Metadata if it already exists.  Even if we
-                    // are indexing `update_metadata`` and there's a new URI, the new background
-                    // task will overwrite it.
+                    asset_data::Column::Metadata,
```
kespinola, line 437: Metadata should not be updated as it would over write what is potentially valid. Still enqueue a download job and if it is the same it will override it.

Metadata should only be set to proccesisng on insert.

kespinola (changes requested):

kespinola (dismissed):

.github/workflows/test.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         run: |
           cargo tree
           git checkout Cargo.lock
-          cargo tree --frozen
+          cargo tree
```
kespinola, line 41: https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/240/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R144

When we update yellowstone-grpc to later package we start getting dependency conflicts. We will be looking at resolving this but would like to do in a follow up.

program_transformers/src/bubblegum/db.rs
```diff
@@ -450,15 +446,23 @@ where
 
     // Do not overwrite changes that happened after decompression (asset_data.base_info_seq = 0).
     // Do not overwrite changes from a later Bubblegum instruction.
+    // Do not update the record if the incoming slot is larger than the current or if it's null.
+    // Update if the current slot on the record is null.
     query.sql = format!(
-        "{} WHERE (asset_data.base_info_seq != 0 AND excluded.base_info_seq >= asset_data.base_info_seq) OR asset_data.base_info_seq IS NULL",
+        "{} WHERE ((asset_data.base_info_seq != 0 AND excluded.base_info_seq >= asset_data.base_info_seq) OR asset_data.base_info_seq IS NULL) AND (excluded.slot_updated <= asset_data.slot_updated OR asset_data.slot_updated IS NULL)",
```
danenbm, line 452: I think this can cause issues with how compressed assets are indexed.

1. I think multiple transactions could occur within the same slot, although it looks like the only two program transformers for bubblegum that change slot_updated are `mint_v1` and `update_metadata` which may be unlikely to be called in the same slot.  But not impossible and people do weird things.

2.  I think the more likely issue is if we ingest transactions out of order, we could be indexing the update_metadata before we index the `mint_v1`, and that would cause maybe some missing data, at first glance owner and delegate.  This maybe could be solved if `update_metadata` writes all the same info as `mint_v1`, but that seems like a risky approach because someone has to remember that or we comment it.

Now, when we decompress the sequence number gets set to 0 so maybe that could be used to control the second condition as well.  But currently I'm not sure whey we need the slot updated condition added now.

grpc-ingest/src/version.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+pub const VERSION: Version = Version {
+    package: env!("CARGO_PKG_NAME"),
+    version: env!("CARGO_PKG_VERSION"),
+    proto: env!("YELLOWSTONE_GRPC_PROTO_VERSION"),
+    solana: env!("SOLANA_SDK_VERSION"),
+    git: env!("GIT_VERSION"),
+    rustc: env!("VERGEN_RUSTC_SEMVER"),
+    buildts: env!("VERGEN_BUILD_TIMESTAMP"),
```
danenbm, line 21: are these supposed to be `VERSION`?

Nagaprasadvr: i got errors changing it

Cargo.toml
```diff
@@ -119,11 +134,16 @@ tokio-stream = "0.1.14"
...
 tower-http = "0.3.5"
 tracing = "0.1.35"
+tracing-opentelemetry = "0.22.0"
 tracing-subscriber = "0.3.16"
 txn_forwarder = { path = "tools/txn_forwarder" }
 url = "2.3.1"
+vergen = "8.2.1"
 wasi = "0.7.0"
 wasm-bindgen = "0.2.83"
+yellowstone-grpc-client = { git = "https://github.com/rpcpool/yellowstone-grpc.git", tag = "v1.15.1+solana.1.18.22" }                          # tag is geyser plugin
```
danenbm, line 144: @kespinola told me these were being moved to https://crates.io/crates/yellowstone-grpc-kafka so could we update these to use that crate?

Nagaprasadvr: It didn't workout with dependencies, so we are doing a new release which resolves this dep issue and update the pr soon

danenbm (commented):
LGTM except for one comment on the crate imports

Cargo.toml
danenbm: Can we make sure all the crates are published versions?

PR #245, opened 2025-02-18
Add asset supply checks and return meaningful error message for getAsset method

Nagaprasadvr:
- [x] check asset supply and return "Asset has no supply" message  instead of asset not found

### DIFF
--- a/digital_asset_types/src/dao/scopes/asset.rs
+++ b/digital_asset_types/src/dao/scopes/asset.rs
@@ -15,7 +15,9 @@ use crate::{
 };
 use indexmap::IndexMap;
 use mpl_token_metadata::accounts::{Edition, MasterEdition};
-use sea_orm::{entity::*, query::*, sea_query::Expr, ConnectionTrait, DbErr, Order};
+use sea_orm::{
+    entity::*, prelude::Decimal, query::*, sea_query::Expr, ConnectionTrait, DbErr, Order,
+};
 use serde::de::DeserializeOwned;
 use serde_json::Value;
 use solana_sdk::pubkey::Pubkey;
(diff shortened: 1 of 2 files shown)


kespinola (approved):

danenbm (approved):
This seems like a good improvement to me adding some others for visibiity.

kespinola:
> This seems like a good improvement to me adding some others for visibility.

Thanks for the review @danenbm for you and others. Triton has been looking closer at the API and will likely update metrics for the API from status = `true | false` to `status_code`. We will keep status to not break current metrics reporting but using code will allow more granularity in reporting to know what was a failed validation vs 404 not indexed.

danenbm (commented):
Ok so looks like this only affects `get_asset`, should a similar thing be done for `get_asset_proof`?

Nagaprasadvr:
> Ok so looks like this only affects `get_asset`, should a similar thing be done for `get_asset_proof`?

yeah i agree, we can add it

danenbm:
I guess since before we hardcoded `include_no_supply` to false it never returned items with no supply and just gave db error.

I have a basic question on how does this work for fungible items vs. NFTs including burned items? 

I guess burned NFTs we still somehow return from `getAsset` even though supply might be zero?

And fungible items we wouldn't return if token supply is 0 anyways?

Nagaprasadvr:
> I guess since before we hardcoded `include_no_supply` to false it never returned items with no supply and just gave db asset not found.
> 
> I have a basic question on how does this work for fungible items vs. NFTs including burned items.
> 
> Burned NFTs we still somehow return from `getAsset` even though supply might be zero because the mint account gets indexed?
> 
> And fungible items we wouldn't return if token supply is 0 anyways?

Burnt assets will have zero supply right

danenbm (approved):

kespinola (approved):

PR #247, opened 2025-02-28
`feat`: handle token program close ixs to delete closed mints and token accounts

Nagaprasadvr:
- [x] handle token program close ixs
- [x] update and delete entries on db accordingly
      deletes - mints, token_accounts
      updates - asset


blockbuster/src/programs/token_account/mod.rs
```diff
@@ -74,4 +76,28 @@ impl ProgramParser for TokenAccountParser {
...
+            inner_ix: _,
+            slot: _,
+        } = bundle;
+
+        if let Some(ix) = instruction {
+            if !ix.data.is_empty() && ix.data[0] == 9 && !keys.is_empty() {
+                return Ok(Box::new(TokenProgramEntity::CloseIx(keys[0])));
+            } else {
+                return Err(BlockbusterError::InstructionTypeNotImplemented);
+            }
```
kespinola, line 98: There is no information in the instruction to distinguish between a mint and token account? Ideally we establish that at this point instead of checking the db for mint and then fallback to TA.

kespinola (approved):

blockbuster/src/programs/token_account/mod.rs
Nagaprasadvr: Instruction data only has the discriminator to map close ix , so we cannot determine what type of account is passed as we only have the pubkey

kespinola (commented):
We need to add integration test.

danenbm (approved):

danenbm (changes requested):
Agree with @kespinola lets add integration test otherwise LGTM

integration_tests/tests/integration_tests/account_close_tests.rs
```diff
@@ -0,0 +1,110 @@
...
+    index_seed_events(&setup, seeds.iter().collect()).await;
+
+    let (mint_res, asset_res) = tokio::join!(
+        tokens::Entity::find_by_id(m_bytes.clone()).one(setup.db.as_ref()),
+        asset::Entity::find_by_id(m_bytes).one(setup.db.as_ref())
+    );
+
+    assert_eq!(mint_res.unwrap().is_none(), true);
+    assert_eq!(asset_res.unwrap().unwrap().burnt, true);
+}
```
kespinola, line 110: You can just use `assert!()` when evaluating truth

kespinola (approved):

PR #249, opened 2025-03-04
Update rust version in toolchain file and docker files

danenbm:
### Docker
I did also try using the lockfile and passing `--locked` to the docker build, but this causes an issue with `cargo install wasm-pack` which still needs the latest Rust version, so just opted to update Rust version.

### Local toolchain file
Updating to latest Rust caused me to need to update time crate as well.

### Testing
Local build works and `docker compose up --force-recreate --build` works.


Nagaprasadvr:
can we use 1.85 ?

Nagaprasadvr (approved):

kespinola (approved):

PR #250, opened 2025-03-13
`impr`: add composite index (owner,amount) and (mint,owner) on token_a…

Nagaprasadvr:
getTokenAccounts calls were kind of slow as we have amount > 0 as a default condition , adding a composite index (owner,amount) and (mint,amount) can fix this potentially and reduce the latency

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -46,6 +46,8 @@ mod m20240520_120101_add_mpl_core_external_plugins_columns;
 mod m20240718_161232_change_supply_columns_to_numeric;
 mod m20241119_060310_add_token_inscription_enum_variant;
 mod m20250213_053451_add_idx_token_accounts_owner;
+mod m20250313_105132_drop_idx_token_accounts_owner_and_idx_ta_mint;
+mod m20250313_105206_add_idx_ta_owner_amount_and_idx_ta_mint_amount;
 
 pub mod model;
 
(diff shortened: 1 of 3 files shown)


kespinola:
A note that DAS doesn't support sort on the `getTokenAccounts` I advocate for DAS API being able to offer the same token query capabilities as RPC.


| RPC Method                     | DAS Equivalent                                         | Status |
|--------------------------------|-------------------------------------------------------|--------|
| `getTokenAccountsByOwner`      | `getTokenAccounts({ params: { owner } })`            | ✅     |
| `getSupply`                    | `getAsset({ id })`                                   | ✅     |
| `getTokenAccountsByDelegate`   | `getTokenAccounts({ params: { delegate } })`        | 🔴     |
| `getTokenLargestAccounts`      | `getTokenAccounts({ params: { owner, options: { sortBy: 'amount', direction: 'DESC' } })` | 🔴     |
| `getTokenAccountBalance`       | `getTokenAccount({ id })`                           | 🔴     |
| `getTokenSupply`               | `getTokenAccount({ id })`                           | 🔴     |

_The `getTokenAccounts` call is missing sorts in DAS right now. With search, it will handle more by being able to page through all tokens._  
_There is no GET endpoint for token account._

kespinola (approved):

danenbm (approved):

PR #251, opened 2025-03-19
Bubblegum Leaf Schema V2

danenbm:
WIP

### DIFF
--- a/digital_asset_types/src/rpc/asset.rs
+++ b/digital_asset_types/src/rpc/asset.rs
@@ -217,6 +217,10 @@ pub struct Compression {
     pub compressed: bool,
     pub data_hash: String,
     pub creator_hash: String,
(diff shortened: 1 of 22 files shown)


digital_asset_types/src/rpc/asset.rs
```diff
@@ -217,6 +217,10 @@ pub struct Compression {
     pub compressed: bool,
     pub data_hash: String,
     pub creator_hash: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub asset_data_hash: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub flags: Option<u8>,
```
andrii-kl, line 223: I would change the name to make it more clear what it is about and what this flag is for.

PR #252, opened 2025-03-20
Bubblegum V2 Instructions

danenbm:
### Add indexing for new Bubblegum instructions
* `burn_v2`
* `delegate_v2`
* `delegate_and_freeze_v2`
* `freeze_v2`
* `mint_v2`
* `set_collection_v2`
* `set_non_transferable_v2`
* `thaw_v2`
* `thaw_and_revoke_v2`
* `transfer_v2`
* `unverify_creator_v2`
* `update_asset_data_v2`
* `update_metadata_v2`
* `verity_creator_v2`

### New Bubblegum functionality
* Bubblegum now has a `LeafSchemaV2` that has `collection_hash`, `asset_data_hash`, and `flags` (which currently hold bits for being frozen or soulbound).
* Added ability to add cNFTs to an mpl-core collection using the V2 instructions, and use permanent delegates on the collection to authorize operations on those cNFTs.
* Added asset-level frozen and non-transferrable checks.
* Added `MetadataArgsV2` which removed unused things like uses and removes concept of collection verified flag.

### Fixed issue
Since the V2 assets use mpl-core collections, the code that automatically updates asset authorities when mpl-core collections are re-indexed (`update_group_asset_authorities`) causes an inconsistent behavior in Bubblegum.

Bubblegum has historically used the `tree_authority` PDA as the asset's authority (which no longer makes a lot of sense given that now the `tree_owner` or collection authority are considered the update authority for Bubblegum assets).  So now that Bubblegum V2 assets are using mpl-core collections, these assets will first have the `tree_authority` as the authority, but when an mpl-core collection is updated, it will switch to collection update authority.  Ideally we'd switch over Bubblegum V2 assets to just use the update authority concept, but we don't know who the authority is at time of mint or update because a delegate could sign.

Thus, to fix this I modified `update_group_asset_authorities` to skip Bubblegum assets.

### Out of scope
* Asset data is on the leaf schema and there's an instruction to update it, but its disabled in the smart contract, so indexing is not needed at this time.  Instruction name is supported, but discarded in blockbuster and program transformer code.

### Testing
* Integration tests added for all Bubblegum V2 instructions.

### Misc
* When I regenerated SeaORM types I picked up a couple other recent changes that I think should have been included in the autogenerated files.
* There's one existing integraiton test `get_token_accounts_by_owner` that's failing that I didn't include an updated snapshot for but I'll take a look when I can.


program_transformers/src/bubblegum/mint.rs
danenbm, line 1: This file was renamed from mint_v1 to mint, that's why its showing as a full new file.

It is mostly same as mint_v1, except I did the leaf schema version matching at the beginning.

digital_asset_types/src/dapi/common/asset.rs
```diff
@@ -479,6 +479,10 @@ pub fn asset_to_rpc(asset: FullAsset, options: &Options) -> Result<RpcAsset, DbE
                 .creator_hash
                 .map(|e| if asset.compressed { e.trim() } else { "" }.to_string())
                 .unwrap_or_default(),
+            asset_data_hash: asset
+                .asset_data_hash
+                .map(|e| if asset.compressed { e.trim() } else { "" }.to_string()),
```
andrii-kl, line 484: Do we really need to use empty String in case of None, since we use skip_serializing_if in the model.

danenbm: I'm just doing it to be consistent with the other hsahes.

PR #266, opened 2025-07-02
Only update assets in mpl-core collection if auth changed

danenbm:
### Notes
Prevents updating the authority for all of the assets in an mpl-core collection for every collection change.

Instead, we update all assets in the collection if:

1. The collection is found and its authority has changed.
2. The collection is not found.

Case 2 requires us to attempt to update all assets in a collection because assets can be indexed before the collection itself.

### Testing
* Created a collection and verified `collection_authority_changed` condition **was** triggered, since collection was not yet indexed.
* Created an asset in the collection and verified `collection_authority_changed` condition **was NOT** triggered for either the collection or asset.
* Updated the collection's update authority and verified `collection_authority_changed` condition **was** triggered and both the collection's authority and asset's authority was updated.
* Updated the collection's name and verified `collection_authority_changed` **was NOT** triggered.

### DIFF
--- a/program_transformers/src/mpl_core_program/v1_asset.rs
+++ b/program_transformers/src/mpl_core_program/v1_asset.rs
@@ -104,8 +104,8 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
         UpdateAuthority::None => Pubkey::default().to_bytes().to_vec(),
     };
 
-    // If we are looking at a collection, see if there's an existing record and if so check
-    // if the authority changed.
+    // If this is a collection, check if there's an existing authority record.
+    // If found, determine whether the authority has changed.
     let collection_authority_changed = match account_data {
         MplCoreAccountData::Collection(_) => {
             let existing_record = asset_authority::Entity::find()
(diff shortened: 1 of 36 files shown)


blockiosaurus (dismissed):

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -104,6 +104,29 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+            // 2. The collection is not found.
+            //
+            // Case 2 requires us to attempt to update all assets in a collection because assets
+            // can be indexed before the collection itself.
+            existing_record
+                .map(|record| record.authority != update_authority)
+                .unwrap_or(true)
+        }
+        _ => false,
+    };
```
kespinola, line 128: I really want to avoid doing lookups while indexing. Is there a way to do this where you join in record you want to check against and then adjusting the where class of the insert or update?

danenbm: i will look into this

PR #268, opened 2025-09-30
Add MIP-11 Support to DAS Indexer

pileks:
## Summary

- Upgrades `mpl-core` dependency to `0.10.1-alpha.2` to support MIP-11 Groups plugin
- Implements indexing for Groups plugin data in the `asset_grouping` table
- Adds logic to track and update asset groupings under the "group" key based on slot ordering
- Clears stale groupings and inserts new ones when Groups plugin is present on assets

## Changes

- Updated `mpl-core` version in `Cargo.toml` and `Cargo.lock`
- Modified `v1_asset.rs` to handle Groups plugin indexing with proper slot-based update logic
- Added `.gitignore` for integration tests `.env` file

## Future work
- Add integration test once core is deployed to devnet


program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -367,6 +366,62 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+        )
+        .one(&txn)
+        .await
+        .map_err(|db_err| ProgramTransformerError::AssetIndexError(db_err.to_string()))?
+        .map(|model| model.slot_updated.unwrap_or(0));
+
+    let last_group_slot_updated = last_group_slot_updated.unwrap_or(0);
+
+    // Only perform updates if the last asset_grouping for group plugin slot_updated is less than the current slot for the asset update.
+    if last_group_slot_updated < slot_i {
```
danenbm: I believe getting `slot_updated` from a separate query and doing the `slot_updated` check outside of the delete/insert can lead to a race condition between concurrent ingester processes.

Example scenario:
1. Process 1, an update from slot 1090 gets `last_group_slot_updated`.
2. Process 2, an update from slot 2512 gets `last_group_slot_updated` and completes its update.
3. Process 1 uses its now stale value of `last_group_slot_updated` and completes its update, effectively downgrading the group entry to an older value.

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -367,6 +366,62 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+                asset_grouping::Column::AssetId
+                    .eq(id_vec.clone())
+                    .and(asset_grouping::Column::GroupKey.eq("group".to_string()))
+                    .and(asset_grouping::Column::SlotUpdated.lt(slot_i)),
+            )
+            .build(DbBackend::Postgres);
+
+        txn.execute(query)
+            .await
+            .map_err(|db_err| ProgramTransformerError::AssetIndexError(db_err.to_string()))?;
```
danenbm: We might want to have the delete and insert for this part of the same txn so if there's a failure they are reverted together

pileks: I have updated the indexer to only fire off a single query for any of the cases - asset without group/empty group just deletes old groupings, while asset with groups now performs deletion and insertion using CTEs.

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -448,6 +447,114 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+                   WHERE asset_id = $1 \
+                     AND group_key = 'group' \
+                     AND (SELECT max_slot FROM max_existing_slot) < $2 \
+                 ) \
+                 INSERT INTO asset_grouping (asset_id, group_key, group_value, verified, group_info_seq, slot_updated) \
+                 SELECT * FROM (VALUES {}) AS v(asset_id, group_key, group_value, verified, group_info_seq, slot_updated) \
+                 WHERE (SELECT max_slot FROM max_existing_slot) < $2 \
+                 ON CONFLICT (asset_id, group_key, group_value) WHERE group_key != 'collection' \
+                 DO NOTHING",
+                values_clause
```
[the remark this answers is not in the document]
pileks: CodeRabbit, are you drunk?

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -448,7 +447,69 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+            verified = EXCLUDED.verified, \
+            group_info_seq = EXCLUDED.group_info_seq \
+            WHERE excluded.slot_updated >= asset_grouping.slot_updated OR asset_grouping.slot_updated IS NULL",
+            query.sql
+        );
+
+    txn.execute(query)
+        .await
+        .map_err(|db_err| ProgramTransformerError::AssetIndexError(db_err.to_string()))?;
+
```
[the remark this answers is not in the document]
pileks: this is intended, as stale groupings get filtered out during queries.

program_transformers/src/mpl_core_program/v1_asset.rs
```diff
@@ -448,7 +447,64 @@ pub async fn save_v1_asset<T: ConnectionTrait + TransactionTrait>(
...
+            "{} ON CONFLICT (asset_id, group_key, group_value) WHERE (group_key != 'collection') DO UPDATE SET \
+            slot_updated = EXCLUDED.slot_updated, \
+            verified = EXCLUDED.verified, \
+            group_info_seq = EXCLUDED.group_info_seq \
+            WHERE excluded.slot_updated >= asset_grouping.slot_updated OR asset_grouping.slot_updated IS NULL",
+            query.sql
+        );
+
+    txn.execute(query)
+        .await
```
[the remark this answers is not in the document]
pileks: We don't consider these same-slot situations a problem currently, as we don't foresee them happening often. If our analysis shows differently, we will adjust.

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -636,6 +643,24 @@ fn filter_out_stale_creators(creators: &mut Vec<asset_creators::Model>) {
...
+    let max_slot_updated = asset_groupings_with_date
+        .iter()
+        .filter(|ag| ag.0.group_key == "group")
+        .map(|ag| ag.0.slot_updated)
+        .max();
+    if let Some(max_slot_updated) = max_slot_updated {
+        asset_groupings_with_date
+            .retain(|ag| ag.0.group_key != "group" || ag.0.slot_updated == max_slot_updated);
+    }
+}
```
danenbm, line 684: Does this need to do any specific handling with the `empty_groups` empty Vec?

danenbm (dismissed):
Change LGTM just had the one question and then also I assume you will want to add testing at some point after its deployed to devnet

digital_asset_types/src/dao/scopes/asset.rs
pileks: You're right - since I'm inserting a grouping with a `null` value, I should filter out that grouping, so that nothing renders in cases where all groups were removed.

danenbm:
Superceded by https://github.com/metaplex-foundation/digital-asset-rpc-infrastructure/pull/270

PR #276, opened 2026-06-17
Inherit SFBP and Creators from collection for bubblegum v2

MarkSackerberg:
Adds read-side support for Bubblegum V2 inherited seller fee basis points (SFBP). When a compressed NFT stores the sentinel value 65535 (0xffff), DAS resolves royalty destinations from the linked MPL Core collection at query time instead of exposing the raw sentinel.

This is intentionally query-layer only: the indexer continues to store royalty_amount = 65535 on the leaf; resolution happens when building the RPC response.

**Background**
Bubblegum V2 soon allows minting into a Core collection with sellerFeeBasisPoints = 65535, meaning “inherit from the collection’s Royalties plugin.” The on-chain leaf stores the sentinel; the effective royalty rate and creator destinations live on the collection.

Previously, DAS would surface basis_points: 65535 (or misleading values) and an empty creators array for inherited mints with no leaf-level creators.

**Changes**
Query-time collection hydration (digital_asset_types)

- Introduces SELLER_FEE_BASIS_POINTS_INHERIT = 65535.
- Extends FullAsset with:
  - inherited_collection_royalty
  - inherited_collection_creators
- Adds hydrate_inherited_sfbp_collection_royalties() for get_by_id and batch asset fetches:
  - Detects Bubblegum V2 assets with royalty_amount == 65535 and a collection grouping
  - Loads royalty + creators from the indexed Core collection row
- Updates asset_to_rpc:
  - basis_points: resolved collection rate (e.g. 750)
  - basis_points_raw: 65535 when inherited
  - royalties.inherited: true when inherited
  - creators: collection Royalties plugin creators when inherited
  - creators_raw: creators on the leaf (should always be empty)
  - Fixes primary_sale_happened mapping (was incorrectly wired to chain data before)

RPC schema (digital_asset_types/src/rpc/asset.rs)

- Optional fields on Royalty:
  - basis_points_raw
  - inherited


nft_ingester/src/main.rs
```diff
@@ -118,17 +118,21 @@ pub async fn main() -> Result<(), IngesterError> {
...
                         bg_task_sender.clone(),
                         ack_sender.clone(),
-                        if i == 0 {
-                            ConsumptionType::Redeliver
-                        } else {
-                            ConsumptionType::New
-                        },
+                        consumption_type,
                         stream_name,
                     );
```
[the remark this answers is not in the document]
MarkSackerberg: Pre-existing on main; this PR only changes consumption_type selection for TXNFILL/ACCFILL

integration_tests/tests/integration_tests/snapshots/integration_tests__cnft_tests_v2_inherited_sfbp__mint_v2_inherited_sfbp_resolves_collection_royalty_and_creators.snap
danenbm, line 1: Shouldn't I see `inherited_collection_royalty` and `inherited_collection_creators` in here?

MarkSackerberg: Copying the Slack message also here:

I had another look at the comment you had for digital-asset-rpc-infrastructure. on FullAsset we have inherited_collection_royalty and inherited_collection_creators in addition to the default fields because i did not want to overwrite the asset level sfbp and creators in there but only did it in the asset_to_rpc function:
```
    let inherited_sfbp = matches!(interface, Interface::MplBubblegumV2)
        && asset.royalty_amount == SELLER_FEE_BASIS_POINTS_INHERIT;
    let royalty_destination_creators = if inherited_sfbp {
        inherited_collection_creators.unwrap_or_default()
    } else {
        creators
    };
```
The idea behind doing it like this was to not overwrite creators + royalties too early in the process and there is also no need to expose the not inherited creators since bubblegum v2 does not allow having creators on the asset itself when the Sentinel is used.
So from my point of view there is no point in adding an always empty creators_raw similar to basis_points_raw (with the sentinel) to the response, but if you would prefer to have it in I could do so

danenbm: I think they key for me is, since the creator array and metadata are the leaf hash that gets derived from DAS, I don't want them to fail because they are hashing inherited creators.  So that's just got to all be clear

solana-test-validator-geyser-config/accountsdb-plugin-config.json
```diff
@@ -17,6 +17,10 @@
     ]
   },
   "transaction_selector": {
-    "mentions": ["BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY"]
+    "mentions": [
+      "BGUMAp9Gq7iTEuizy4pqaxsTyUCBK68MDfK752saRPUY",
+      "CoREENxT6tW1HoK8ypY1SxRMZTcVPm7R94rH4PZNhX7d",
+      "mcmt6YrQEMKw8Mw43FmpRLmf7BqRnFMKmAcbxE3xkAW"
```
danenbm: Wondering if we need these extra mentions, because we only parse Bubblegum-related transactions.

nft_ingester/src/main.rs
```diff
@@ -118,17 +118,20 @@ pub async fn main() -> Result<(), IngesterError> {
...
                         database_pool.clone(),
                         config.get_messneger_client_config(),
                         bg_task_sender.clone(),
                         ack_sender.clone(),
-                        if i == 0 {
-                            ConsumptionType::Redeliver
-                        } else {
-                            ConsumptionType::New
-                        },
+                        consumption_type,
```
danenbm: Why do we need these changes?  Seems like we shouldn't mess with this for this PR, could keep it out or move to separate PR?

digital_asset_types/src/rpc/asset.rs
```diff
@@ -297,6 +297,10 @@ pub struct Royalty {
     pub target: Option<String>,
     pub percent: f64,
     pub basis_points: u32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub basis_points_inherited: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub percent_inherited: Option<f64>,
```
danenbm: Ultra nit: Could these just be in the same order as above, so:

percent
basis_points
percent_inherited
basis_points_inherited

or

percent
percent_inherited
basis_points
basis_points_inherited

I prefer the first one

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -437,8 +441,10 @@ pub async fn get_related_for_assets(
     for (_id, asset) in assets_map.iter_mut() {
         filter_out_stale_asset_groupings(&mut asset.groups);
     }
+    let mut assets: Vec<FullAsset> = assets_map.into_values().collect();
+    hydrate_inherited_sfbp_collection_royalties(conn, &mut assets).await?;
```
danenbm, line 445: Every test that reaches hydration (the two mock-DB unit tests and the integration scenario) goes through get_by_id, so hydrate_inherited_sfbp_collection_royalties only ever runs on a single-element slice — the multi-asset bookkeeping (index mapping, collection-id dedup, fan-out) and the get_related_for_assets call site used by searchAssets/getAssetsByOwner/getAssetBatch are never executed.

Could you add an integration test that exercises this through a list endpoint, e.g. getAssetsByOwner for the fixture owner? The existing fixtures should already support that. Even better if the fixture script mints a second sentinel cNFT to the same owner in the same collection — that would also cover the dedup path (two assets resolving through one collection lookup).

digital_asset_types/src/dao/scopes/asset.rs
```diff
@@ -706,6 +718,72 @@ fn filter_out_stale_asset_groupings(
...
+    conn: &impl ConnectionTrait,
+    assets: &mut [FullAsset],
+) -> Result<(), DbErr> {
+    let mut asset_to_collection = Vec::new();
+    let mut collection_ids = Vec::new();
+
+    for (i, asset) in assets.iter().enumerate() {
+        if asset.asset.royalty_amount != SELLER_FEE_BASIS_POINTS_INHERIT {
+            continue;
+        }
```
danenbm, line 731: Consider also gating on specification_asset_class == MplBubblegumV2 (mirroring the interface gate in asset_to_rpc) and on the grouping being verified. Not exploitable today — the program guarantees sentinel ⇒ verified V2 collection (including blocking SetCollectionV2 removal via CannotRemoveFromCollectionWithInheritedSellerFee) — but this makes the read path locally self-evident instead of depending on write-path invariants across the program, blockbuster, and the transformer, and skips two wasted queries if junk rows ever exist.

danenbm (changes requested):

nft_ingester/src/main.rs
MarkSackerberg: agree

digital_asset_types/src/dao/scopes/asset.rs
MarkSackerberg: Added test_mint_v2_inherited_sfbp_resolves_via_get_assets_by_owner

digital_asset_types/src/dao/scopes/asset.rs
MarkSackerberg: Done

MarkSackerberg:
@danenbm updated / reverted as discussed. Please rereview

danenbm (approved):
LGTM.  Thank you for going back and forth on the design.  I think we landed on the right choice.   Good tightening around the hydration stuff, and also well tested PR overall.

PR #277, opened 2026-08-31
Add Solana transaction v1 support

nhanphan:
## Summary

- upgrade the workspace to the Agave/Solana 4.2 client and transaction types required for transaction v1
- request transaction version 1 with base64 encoding in RPC, backfill, forwarding, and integration-test paths
- add a v1-aware Plerkle serializer and bridge the legacy Plerkle wire schema into current Solana instruction types
- preserve Token-2022 extension output compatibility across the dependency upgrade
- test a real v1 transaction whose serialized size exceeds the legacy 1,232-byte limit

## Validation

- `cargo check --workspace --all-targets`
- `cargo test --workspace --lib` (11 passed)
- `cargo clippy --all-targets --tests -- -Dwarnings`
- `cargo fmt --all -- --check`
- `cargo tree --frozen`
- `git diff --check`

## External validation boundary

The newest published `plerkle-test-validator` image is still based on Agave 3.1.13. The in-repo RPC, backfill, wire-consumer, and parser paths are v1-ready, but live Geyser v1 emission cannot be exercised until an Agave 4.2-compatible Plerkle plugin image is published.

PR #278, opened 2026-09-16
Add Solana transaction v1 support

ArthurPaivaT:
Adds support for Solana transaction v1

  - Moves the workspace to the Solana 4.2.2 crates and Rust 1.96.1, which are needed to decode v1 transactions.
  - Requests transaction version 1 with base64 encoding everywhere DAS pulls transactions from the RPC.
  - Adds a v1-aware Plerkle serializer, because `plerkle_serialization` 3.0.0 can't decode v1.
  - Returns errors instead of panicking when the RPC returns account data that can't be decoded.
  - Tests a synthetic v1 transaction and a real 3726-byte devnet v1 transaction, from RPC bytes to the Plerkle stream.
  - Built on top of #277 by @nhanphan.


blockbuster/src/programs/token_extensions/extension.rs
```diff
@@ -232,23 +233,23 @@ impl From<MemoTransfer> for ShadowMemoTransfer {
 impl From<MetadataPointer> for ShadowMetadataPointer {
     fn from(original: MetadataPointer) -> Self {
         ShadowMetadataPointer {
-            metadata_address: original.metadata_address,
+            metadata_address: bytemuck::cast(original.metadata_address),
```
danenbm, line 236: Note for reviewers on the `bytemuck::cast` calls throughout this file: these are purely mechanical fallout from the Solana 4.2 dependency bump, no behavior change.

The shadow structs here declare fields like `OptionalNonZeroPubkey` from the `spl_pod` version this workspace depends on directly, while the extension structs coming out of the newer `spl-token-2022` expose the same-named types built against a different `solana-pubkey` generation (the lockfile now has `solana-pubkey` 2.4.0, 3.0.0, and 4.2.0 coexisting). Rust treats those as distinct types even though they are byte-identical wrappers, so the old plain field assignments stopped compiling.

`bytemuck::cast` is a compile-time size-checked byte reinterpretation between `Pod` types — if the layouts ever diverged it would fail to compile rather than corrupt data, and it is zero-cost at runtime. Related: the pod encryption types (`PodElGamalPubkey`, `PodAeCiphertext`, etc.) moved out of `spl_token_2022::solana_zk_sdk` into the standalone `solana-zk-sdk-pod` crate, hence the new import/dependency. Once the ecosystem converges on a single `solana-pubkey` version these casts can be deleted and plain assignment will work again.

danenbm (approved):
Thank you for this excellent contribution!

Reviewed the full diff and ran the test suite locally — everything passes, and the workspace builds cleanly on the updated toolchain.

Also independently verified the behavior end to end against a live devnet DAS endpoint running this code: sent Bubblegum operations (create tree, mint, transfer, delegate, burn) plus SPL Token / Token-2022 close-account flows inside Solana **v1-format transactions** (SIMD-0385), and all of it indexed correctly.

Notes from the review:
- The local `serialize_encoded_transaction_with_status` in `das-core` is a clean bridge while the published `plerkle_serialization` crate remains pinned to Solana 3.x types. Mapping `VersionedMessage::V1` to flatbuffer version value `2` is sensible; the read path in `nft_ingester/src/plerkle.rs` is version-agnostic so it handles all formats uniformly.
- Converting the account-decode `.expect()` panics into `ClientError`s in `solana_rpc.rs` is a nice hardening win, and moving decode failures outside the retry loop is the right call since retrying a decode failure cannot help.
- The regression test with a real oversized (>1232 byte) devnet v1 transaction fixture is great coverage.

Follow-on work (tracked separately, not blocking this PR): update `digital-asset-validator-plugin` so Geyser/plerkle natively support transaction v1 — i.e., migrate it to agave 4.2.x, add the `V1` match arms emitting the same version value `2` (ideally adding it to the `.fbs` schema), and publish new `plerkle_serialization` + plugin releases. Once that ships, DAS can consume the published crate and drop the local serializer copy. The one coordination point is keeping the version enum value `2` consistent between the two repos.

